PR #1, opened 2017-07-19
Adds CNAME under /docs

itamarhaber:
Required so pushing the gh pages does not clear the custom domain name.


itamarhaber:
@mnunberg I have merge powers here, but it is your baby after all so you get the honors :)

PR #11, opened 2017-08-17
Commands.md: Clarify BF.RESERVE's error_rate arg

Deimos:
The documentation related to BF.RESERVE's error_rate arg was a bit
misleading and confusing. Specifically, it said that greater values
would result in more space consumption when the opposite is true, and
described it as a "percentage" when it's a ratio, which might result in
people choosing values outside the correct (0, 1) range.

### DIFF
--- a/docs/Commands.md
+++ b/docs/Commands.md
@@ -10,22 +10,24 @@ BF.RESERVE {key} {error_rate} {size}
 
 ### Description:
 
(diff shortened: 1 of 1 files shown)


dvirsky:
On top of that we should reject invalid values with a proper error @mnunberg 

_Sent from my OnePlus ONE A2001 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

PR #13, opened 2017-08-23
Chunked save

mnunberg:


mnunberg:
I know it's missing documentation, but I'm not going to bother writing docs if the API is bad :)

src/rebloom.c
```diff
@@ -220,6 +221,86 @@ static int BFInfo_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv, in
     return REDISMODULE_OK;
 }
 
+#define MAX_SCANDUMP_SIZE 10485760 // 10MB
+
+/**
+ * BF.SCANDUMP <ITER>
```
dvirsky: should be <key> <iter>

PR #24, opened 2017-12-26
Cuckoo filter

mnunberg:
Folks - this all passes tests, etc. (I might want to tweak performance, but no leaks and it seems 'Ok' for now).

let me know if all the hash magic etc. makes sense.


itamarhaber:
@mnunberg I'll try reviewing but I'm a little swamped atm.

PR #31, opened 2018-03-07
fix bloomfilter aof rewrite bugs.

chenyang8094:


mnunberg:
We're going to end up removing AOF rewrite support (in favor of newer 'rdb preamble' AOF support).

PR #33, opened 2018-03-31
Hash64

arinkverma:


mnunberg:
I had something similar in mind. The problem is trying to make this work with older versions. In other words, the module would need to be able to read both 32 and 64 bit versions of the hash.

mnunberg:
Furthermore there is an alternative branch which already contains a conversion to 64 bit hashes.

PR #70, opened 2019-06-06
Top k

ashtul:
Small changes and bugs fix -
1. Bloom now support make coverage which runs all tests and prints out coverage summery
2. Some markdown changes
3. TopK.List now returns nil for empty HeapBuckets
4. Fixed saving to rdb bug
5. Expanded python tests


ashtul:
Now ready for review

src/rm_topk.c
```diff
@@ -197,6 +197,9 @@ void TopKRdbSave(RedisModuleIO *io, void *obj) {
                                  topk->width * topk->depth * sizeof(Bucket));
     RedisModule_SaveStringBuffer(io, (const char *)topk->heap,
                                  topk->k * sizeof(HeapBucket));
+    for(uint32_t i = 0; i < topk->k; ++i) {
+        RedisModule_SaveStringBuffer(io, topk->heap[i].item, strlen(topk->heap[i].item) + 1);
```
jeffreylovitz: This assumes that the heap is full, and will crash when the items are NULL.

```
redis-cli TOPK.RESERVE test 3 5 4 0.9
redis-cli save
```
Will segfault here, since you're calling `strlen` on a NULL value.

src/rm_topk.c
```diff
@@ -148,14 +148,14 @@ int TopK_List_Cmd(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
...
-        if(heapList[i]) {
+    uint32_t k = topk->k;
+    char **heapList = TOPK_CALLOC(k, (sizeof(char *)));
+    TopK_List(topk, heapList);
+    RedisModule_ReplyWithArray(ctx, k);
+    for(int i = 0; i < k; ++i) {
+        if(heapList[i] != NULL) {
             RedisModule_ReplyWithSimpleString(ctx, heapList[i]);
-        }
+        } else RedisModule_ReplyWithNull(ctx);
```
jeffreylovitz: I wouldn't mix curly braces and naked statements in one if/else block (personally I favor always using curly braces, but to each their own).

src/topk.c
```diff
@@ -167,26 +167,23 @@ size_t TopK_Count(TopK *topk, const char *item, size_t itemlen) {
     Bucket *runner = NULL;
     uint32_t fp = TOPK_HASH(item, itemlen, GA);    
     // TODO: The optimization of >heapMin should be revisited for performance
-    counter_t heapMin = checkExistInHeap(topk, item, itemlen) ? topk->heap->count : 0;
+    counter_t heapMin = topk->heap->count;
+    HeapBucket *heapPtr = checkExistInHeap(topk, item, itemlen) ? topk->heap->count : 0;
```
jeffreylovitz: Compiler warning:
`warning: initialization makes pointer from integer without a cast`

What is supposed to be happening here?

ashtul: I'm using the pointer as a flag to know if number exist in heap

ashtul: Great point. just saving the point and later checking whether it is NULL.

src/rm_topk.c
ashtul: Done

src/rm_topk.c
ashtul: Fixed

ashtul:
@jeffreylovitz - worked on your excellent comments.
Please review again

src/rm_topk.c
```diff
@@ -198,7 +200,12 @@ void TopKRdbSave(RedisModuleIO *io, void *obj) {
     RedisModule_SaveStringBuffer(io, (const char *)topk->heap,
                                  topk->k * sizeof(HeapBucket));
     for(uint32_t i = 0; i < topk->k; ++i) {
-        RedisModule_SaveStringBuffer(io, topk->heap[i].item, strlen(topk->heap[i].item) + 1);
+        if(topk->heap[i].item != NULL) {
+            RedisModule_SaveStringBuffer(io, topk->heap[i].item, strlen(topk->heap[i].item) + 1);
+        } else {
+            char *emptystr = "\n";
+            RedisModule_SaveStringBuffer(io, emptystr, 1);
```
jeffreylovitz: Serialize an empty string like:
`RedisModule_SaveStringBuffer(io, "", 1);`
Which causes the string to consist of a single NULL byte. The method here has 2 bytes, so serialization still turns a bit odd:

```
$ redis-cli TOPK.RESERVE test 3 5 4 0.9
OK
$ redis-cli TOPK.LIST test
1) (nil)
2) (nil)
3) (nil)
$ redis-cli debug reload
OK
$ redis-cli TOPK.LIST test
1)

2)

3)
#
```

ashtul: Solved

PR #80, opened 2019-06-20
Heavy hitter detection

ashtul:
Major change -
TOPK.ADD command usually returns null unless an item gets into the list. Then, the dropped item is returned.

Bug fix -
topk.c line 180 - TOPK.QUERY didn't return items equal to minimum item.


src/rm_topk.c
```diff
@@ -77,28 +77,37 @@ int TopK_Create_Cmd(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
...
+    for(int i = 0; i < itemCount; ++i) {
+        size_t itemlen;
+        const char *item = RedisModule_StringPtrLen(argv[i + 2], &itemlen);
+        char *expelledItem = TopK_Add(topk, item, itemlen);
+  
+        if(expelledItem == NULL) {
+            RedisModule_ReplyWithNull(ctx);
+        } else {
+            RedisModule_ReplyWithSimpleString(ctx, expelledItem);
+            free(expelledItem);
```
jeffreylovitz: This was formerly a call to `TOPK_FREE` (`topk.c:148`).

ashtul: @jeffreylovitz , great catch. Thank you.

PR #83, opened 2019-06-24
Added python client to docs

ashtul:


K-Jo:
We have spoken about this.  The client documentation should belong in the driver's repo's.  The documentation in here should be commands only.

K-Jo:
https://oss.redislabs.com/redisbloom/Java_Client/ should also move to the java client

ashtul:
@K-Jo , not sure what do you mean then. Should I remove these two stabs all together?

PR #86, opened 2019-06-26
Coverage badge using DoceCov

ashtul:


ashtul:
Message on the current missing badge is "No commits found on master yet." 
So I expected once we merge it, it will show the stats.

gkorland:
@ashtul thanks! looks good, I guess we just need to add more tests now...

ashtul:
@gkorland some are checks we don't want to spend the time emulating.
Kudos to @rafie for helping out.

gkorland:
@ashtul  we need to cover error paths too.
but if there are paths you think are not related to the module we should mark them for ignore

PR #87, opened 2019-06-27
Expand coverage

ashtul:


src/cuckoo.c
```diff
@@ -374,6 +374,7 @@ size_t CuckooFilter_Compact(CuckooFilter *cf) {
     return ret;
 }
 
+/* CF.DEBUG uses another function
```
gkorland: @ashtul why do you keep dead code?

src/rm_cms.c
```diff
@@ -34,9 +34,9 @@ static int GetCMSKey(RedisModuleCtx *ctx, RedisModuleString *keyName, CMSketch *
 
 static int CreateCMSKey(RedisModuleCtx *ctx, RedisModuleString *keyName, long long width,
                         long long depth, CMSketch **cms, RedisModuleKey **key) {
-    if (*key == NULL) {
+    /*if (*key == NULL) {
```
gkorland: @ashtul why do you keep dead code?

ashtul: Made some more cleanup.

src/cuckoo.c
ashtul: It is an info function which doesn't hurt to keep for possible future use.

PR #98, opened 2019-07-15
Hash fix

ashtul:
fixed linker flags
changed function name for future proof

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ LD=gcc
 # Compile flags for linux / osx
 ifeq ($(uname_S),Linux)
 	SHOBJ_CFLAGS ?=  -fno-common -g -ggdb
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -13,7 +13,7 @@ LD=gcc
 # Compile flags for linux / osx
 ifeq ($(uname_S),Linux)
 	SHOBJ_CFLAGS ?=  -fno-common -g -ggdb
-	SHOBJ_LDFLAGS ?= -shared -Bsymbolic -Bsymbolic-functions
+	SHOBJ_LDFLAGS ?= -shared -Wl,-Bsymbolic,-Bsymbolic-functions
```
rafie: I think -Bsymbolic-functions is redundant, but if you wanna keep it you need to introduce it with it's own -Wl.

PR #101, opened 2019-07-16
Hash fix include fix for MacOS

ashtul:
Hopefully


ashtul:
Hi @mnunberg, this is a pretty important one and should take a minute.

PR #102, opened 2019-07-17
Fixed documentation on BF.INSERT.

nicholasamorim:
Fixed example on docs where it states that the command inserts one item but the example provided actually inserts two items.


ashtul:
@nicholasamorim please redirect PR to https://github.com/RedisBloom/RedisBloom/tree/Doc-fix.
We don't merge directly into master.

PR #105, opened 2019-07-29
Fixed link to License file

amiramm:
Old link was to a /docs license file which didn't exist and produced a 404.
Now linking to License file under root, the same way other Redis module repositories do.


amiramm:
@K-Jo - can you review?

PR #109, opened 2019-08-23
Add CF.RESERVE example

dnielsen:
add CF.RESERVE example and changed key name from newFilter to newCuckooFilter

### DIFF
--- a/docs/Quick_Start.md
+++ b/docs/Quick_Start.md
@@ -71,10 +71,18 @@ OK
 
 ## Cuckoo: Adding new items to a filter
 
+
+> Create an empty cuckoo filter with an initial capacity (of 1000 items)
+
+```
+127.0.0.1:6379> CF.RESERVE newCuckooFilter 1000
+(integer) 1
+```
+
 > A new filter is created for you if it does not yet exist
 
 ```
-127.0.0.1:6379> CF.ADD newFilter foo
+127.0.0.1:6379> CF.ADD newCuckooFilter foo
 (integer) 1
 ```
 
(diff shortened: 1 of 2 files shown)


K-Jo:
@dnielsen I would leave this as is and stick to the error messages all redis users are used to.

K-Jo:
@dnielsen can you please remove the line with the change of error message so we can progress?

docs/Quick_Start.md
```diff
@@ -83,18 +91,18 @@ You can add the item multiple times. The filter will attempt to count it.
...
-127.0.0.1:6379> CF.EXISTS newFilter notpresent
+127.0.0.1:6379> CF.EXISTS newCuckooFilter notpresent
 (integer) 0
 ```
 
 ## Cuckoo: Deleting item from filter
 
 ```
-127.0.0.1:6379> CF.DEL newFilter foo
+127.0.0.1:6379> CF.DEL newCuckooFilter foo
```
ashtul: We don't currently have CF.DEL. Did you mean to write DEL?

src/redismodule.h
```diff
@@ -63,7 +63,7 @@
 #define REDISMODULE_HASH_DELETE ((RedisModuleString*)(long)1)
 
 /* Error messages. */
-#define REDISMODULE_ERRORMSG_WRONGTYPE "WRONGTYPE Operation against a key holding the wrong kind of value"
+#define REDISMODULE_ERRORMSG_WRONGTYPE "WRONGTYPE Operation against a key holding the wrong kind of value or key already exists"
```
ashtul: We should not change this but rather separate the cases and return the appropriate message for each one of them. Please remove.

PR #110, opened 2019-08-29
Add CF.RESERVE example (#109)

ashtul:
* Add CF.RESERVE example
add CF.RESERVE example and changed key name from newFilter to newCuckooFilter


ashtul:
Getting #109 into master

ashtul:
@K-Jo please merge

PR #111, opened 2019-09-03
Fixing (most) pedantic warnings

ashtul:


src/cms.c
```diff
@@ -54,7 +54,7 @@ void CMS_IncrBy(CMSketch *cms, const char *item, size_t itemlen, size_t value) {
 
 size_t CMS_Query(CMSketch *cms, const char *item, size_t itemlen) {
     assert(cms);
-    assert(item >= 0);
+    assert(item);
```
gkorland: @ashtul 
    assert(item >= 0);	 !=  assert(item);

ashtul: @gkorland 
item >= 0 is meaningless since pointers are always >=0

PR #117, opened 2019-09-12
add and use RMUtil_RegisterWriteDenyOOMCmd

ashtul:


gkorland:
@ashtul what about the other methods? BF.* and CF.*?
It seems like we might even overuse it there 
e.g.
https://github.com/RedisBloom/RedisBloom/blob/d86e8317465753a09cf4c6d9e15b70df254d8fb1/src/rebloom.c#L1007

ashtul:
The filter might expand if it reaches a certain threshold so it should still use OOM.

ashtul:
Add might locate memory if current filter reaches a certain fill rate.

Issue #46, opened 2018-07-30
Don't replicate unless command is successful.


mnunberg:

gkorland:
@mnunberg @ashtul is that issue still relevant?

ashtul:
#121

PR #121, opened 2019-09-22
moved and removed ReplicateVerbatim func

ashtul:
#46 @gkorland @mnunberg 
* Moved ReplicateVerbatim to bottom of relevant functions.
* Made ReplicateVerbatim common between add/insert function and after additional checks in internal function.
* Removed ReplicateVerbatim from CFcheck function.
* Not sure about leaving ReplicateVerbatim in LoadChunk function.


src/rebloom.c
```diff
@@ -419,14 +414,14 @@ static int BFLoadChunk_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **arg
     if (SBChain_LoadEncodedChunk(sb, iter, buf, bufLen, &errMsg) != 0) {
         return RedisModule_ReplyWithError(ctx, errMsg);
     } else {
+        RedisModule_ReplicateVerbatim(ctx); // Should be replicated?
```
jeffreylovitz: I don't know the answer to this comment! Additionally, we can return with an OK status at 405 - does replication need to occur there?

ashtul: @gkorland @itamarhaber 
for loading chunks, should we replicate? Do we assume the load is done for the whole cluster?

ashtul:
@jeffreylovitz 
This is an excellent question which I guess can be relevant for anywhere we are using rand() in the code to change data.
I guess we can use the first char as a seed? What do you think?

PR #125, opened 2019-10-05
add seed for rand()

ashtul:
@jeffreylovitz 
Use the item's hash as a seed for rand() to stabilize function results across different shards.


ashtul:
Cancel PR
It won't completely eliminate diffs between shards due to inconsistent order of items appearance.
Full randomness is better for ```chance = rand() ```.

PR #128, opened 2019-10-09
cover CFMaxExpansion

ashtul:


ashtul:
@gkorland added documentation

ashtul:
Seems failure is because of different distribution on filters which results in different filter numbers 16 vs 17).
`Traceback (most recent call last):
  File "cuckoo.py", line 197, in test_bucket_size
    self.assertEqual(self.cmd('CF.DEBUG c'), 'bktsize:4 buckets:16 items:1000 deletes:0 filters:16 max_iterations:500')
AssertionError: 'bktsize:4 buckets:16 items:1000 deletes:0 filters:17 max_iterations:500' != 'bktsize:4 buckets:16 items:1000 deletes:0 filters:16 max_iterations:500'`

PR #129, opened 2019-10-12
comment out checked condition - LGTM

ashtul:


contrib/bloom.c
```diff
@@ -130,7 +130,7 @@ int bloom_init(struct bloom *bloom, unsigned entries, double error, unsigned opt
 
     if (options & BLOOM_OPT_ENTS_IS_BITS) {
         // Size is determined by the number of bits
-        if (entries == 0 || entries > 64) {
+        if (/* entries == 0 || */ entries > 64) {
```
gkorland: @ashtul why leave dead code?

ashtul: @gkorland 
Because it is part of the algorithm. If someone reads it, it clearer.
It doesn't harm performance and my sense is, it improves readability is better.

PR #136, opened 2019-10-28
Expose cf.compact in api

ashtul:


src/rebloom.c
```diff
@@ -677,6 +677,24 @@ static int CFDel_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int
     return RedisModule_ReplyWithLongLong(ctx, CuckooFilter_Delete(cf, hash));
 }
 
+static int CFCompact_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
+    RedisModule_AutoMemory(ctx);
+    RedisModule_ReplicateVerbatim(ctx);
```
rafie: Move to end of function? (a-la #121)

src/rebloom.c
```diff
@@ -677,6 +677,24 @@ static int CFDel_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int
...
+
+    if (argc != 2) {
+        return RedisModule_WrongArity(ctx);
+    }
+
+    RedisModuleKey *key = RedisModule_OpenKey(ctx, argv[1], REDISMODULE_READ | REDISMODULE_WRITE);
```
rafie: This key comes from argv[1]. Shouldn't we check for NULL?

rafie:
Not strictly related to this PR, but wouldn't it be a good idea to use Bloom Python client in the unit tests?

src/rebloom.c
ashtul: No, since we checked before we have two arguments

PR #137, opened 2019-10-30
Error-rate and fill-rate benchmarks

ashtul:


filipecosta90:
@ashtul suggest to add the benchmark as use-cases for #292

PR #138, opened 2019-10-31
SubFilter & configurable expansion

ashtul:
Individual sub filters got a struct.
Expansion option allows configuring the ratio at which sub filters will grow.

### DIFF
--- a/src/cuckoo.c
+++ b/src/cuckoo.c
@@ -30,11 +30,8 @@ static size_t getNextN2(size_t n) {
 }
 
 int CuckooFilter_Init(CuckooFilter *filter, size_t capacity, uint16_t bucketSize, uint16_t maxIterations, uint16_t expansion) {
-    //assert(capacity >= bucketSize);
-    assert(isPower2(expansion));
-
     memset(filter, 0, sizeof(*filter));
-    filter->expansion = expansion;
+    filter->expansion = getNextN2(expansion);
     filter->bucketSize = bucketSize;
     filter->maxIterations = maxIterations;
     filter->numBuckets = getNextN2(capacity / bucketSize);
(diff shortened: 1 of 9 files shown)


src/rebloom.c
```diff
@@ -677,15 +695,6 @@ static int CFDel_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv, int
     return RedisModule_ReplyWithLongLong(ctx, CuckooFilter_Delete(cf, hash));
 }
 
-static void fillCFHeader(CFHeader *header, const CuckooFilter *cf) {
```
ashtul: Moved to cf.c with CFHeader_Load

ashtul:
@mnunberg 
For the most part, the changes are to support the new subCF struct.

Logic differences:

- Max iteration is now 20. 500 didn't improve the fill rate.
- We don't randomize the bucket from which we take victim but go 1,2... and then roll back if we need to create a new filter. With the exponential expansion, we need a larger location stamp which we lost so our only option is to roll back and insert the latest item into the new filter and not the victim.

src/cuckoo.c
```diff
@@ -220,89 +253,93 @@ static uint8_t *Filter_FindUnique(CuckooBucket bucket, size_t index, uint16_t bu
...
+                                          const LookupParams *params);
 
 static CuckooInsertStatus CuckooFilter_InsertFP(CuckooFilter *filter, const LookupParams *params) {
-    CuckooBucket *curFilter = filter->filters[filter->numFilters - 1];
-    uint8_t *slot = Filter_FindAvailable(curFilter, filter->bucketSize, params);
-    if (slot) {
-        *slot = params->fp;
-        filter->numItems++;
-        return CuckooInsert_Inserted;
+    for(size_t ii = filter->numFilters; ii > 0; --ii) {
```
mnunberg: Is this strictly needed for this change? I guess if we 'promote' to a larger filter, then that filter out to get used more, rather than starting from the first filter..

ashtul: @mnunberg 
The thinking here was, a filled slot in a small filter contributes more to the error rate than a filled slot in a large filter. Therefore, if we open a larger filter, we should attempt to use it first.

PR #143, opened 2019-11-09
use RMUtil_DefaultAofRewrite not custom functions

ashtul:


ashtul:
Generic should not be used as LOADCHUNK is required

PR #144, opened 2019-11-11
Fix a typo (WEIGHT instead of WEIGHTS)

palicao:
Fix a typo (WEIGHT instead of WEIGHTS) in CountMinSketch docs


ashtul:
Thank you @palicao

PR #149, opened 2019-11-13
Bf/cf.info

ashtul:
- Done cf.info
- For bf.info, should consider what info should be included

### DIFF
--- a/docs/Bloom_Commands.md
+++ b/docs/Bloom_Commands.md
@@ -296,3 +296,33 @@ O(log N), where N is the number of stacked filters in the data structure.
 ### Returns
 
 `OK` on success, or an error on failure.
+
+```
+BF.INFO {key}
+```
+
+### Description
+
+Return information about `key`
+
+### Parameters
+
+* **key** Name of the key to restore
+
+### Complexity O
+
+O(1)
+
+### Returns
+
+```sql
+127.0.0.1:6379> BF.INFO cf
+1) Capacity
+2) (integer) 1709
+3) Size
+4) (integer) 2200
+5) Number of filters
+6) (integer) 1
+7) Number of items inserted
+8) (integer) 0
+```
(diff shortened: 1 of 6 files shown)


jerroldobo:
what unit for the size? Is there any calculator that can help me understand relationship between capacity and size? I tried BF.INFO and the number does not make sense to me in the moment. Since the size is much smaller than I expected.

ashtul:
@jerroldobo The number is bytes of memory.
You also have some overhead of the structs themselves but with large filters this is insignificant.
Please note that each sub-filter uses an extra hash which yields for additional storage space.
If you want to write me an example of a configuration I can try and explain it.

jerroldobo:
@ashtul Thanks for replying, I tried myself and wrote some data in to Bloom filter, for 100M capacity with 0.00001  error rate and 3M items insertion, it seems to me that each data is about 3.8 bytes. I tried to estimate how much size Redis I needed and how the item size would affect error rate. In addition, is https://hur.st/bloomfilter/?n=100M&p=1.0E-6&m=&k= still valid in Redisbloom?

ashtul:
@jerroldobo 
Your error rate is 1/100K (vs. 1/1M in the link you provided) which means k=17 (or k=20 in the link). **Each hash requires 1.44 bits** so with k=20 you get to ~29 bits per item or ~3.8 bytes. If 1/1K is enough for you, you will be using half the amount of memory since k=10.

The number of items you have inserted is not relevant because the full amount of memory is allocated when the filter is created. If you don't insert all the items you prepared the filter for, your error rate will be lower than intended but you will not gain back any memory.

jerroldobo:
Sorry for missing a 0 there and thanks for detailed explanation. I am creating an alert for in case inserted items more than bloom filter capacity is. So I am wondering how the error rate growth after insertion reaches capacity. But after second thought, I think it might be more plausible to set alert before it reaches capacity. However, I still think it might be more clear if we know some tech details of what is the difference between RedisBloom and Bloom filter in theory.

PR #151, opened 2019-11-13
Fix bloom chunk test

ashtul:
A long test which will enough data to create several large filters and check chunking them up.


tests/pytests.py
```diff
@@ -94,17 +94,22 @@ def test_rdb_reload(self):
...
+        self.cmd('bf.reserve', 'myBloom', error_rate, quantity *1024 * 8)
         # test is probabilistic and might fail. It is OK to change variables if 
         # certain to not break anything
-        def do_verify():
-            for x in xrange(1000):
-                self.cmd('bf.add', 'myBloom', x)
+        def do_verify(add):
+            false_positives = 0.0
+            for x in xrange(quantity):
+                if add is True:
```
rafie, line 105: if add:

PR #153, opened 2019-11-14
Bloom expansion config

ashtul:


src/rebloom.c
```diff
@@ -129,6 +130,15 @@ static int BFReserve_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv,
         return RedisModule_ReplyWithError(ctx, "ERR bad capacity");
     }
 
+    long long expansion = BF_DEFAULT_EXPANSION;
+    int ex_loc = RMUtil_ArgIndex("EXPANSION", argv, argc);    
+    if (ex_loc != -1) {
+        if (RedisModule_StringToLongLong(argv[ex_loc + 1], &expansion) != REDISMODULE_OK) {
+            RedisModule_ReplyWithError(ctx, "ERR bad expansion");
```
rafie: Should be:
```
return RedisModule_ReplyWithError(ctx, "ERR bad expansion");
```
Contrary to correct intuition, methods should't return REDISMODULE_ERR when faced with an error. Thus, RedisModule_ReplyWithError returns REDISMODULE_OK.

ashtul: @rafie 
You are correct and it is all over the file.
Talking this opportunity to fix this.

docs/Bloom_Commands.md
```diff
@@ -34,6 +34,13 @@ filter entry.
     been exceeded. Performance will degrade linearly as the number of entries
     grow exponentially.
 
+Optional parameters:
+
+* **expansion**: When a new filter is created, its size will be the size of the
```
K-Jo: Can you elaborate on the when?  I believe it's not really when it's created, but "when a threshold is reached" ... "an additional filter is created which size ...".

docs/Bloom_Commands.md
```diff
@@ -34,6 +34,13 @@ filter entry.
...
+current filter multiplied by `expansion`.
+Default expansion value is 2. New sub-filter size is (2 * expansion) of the
```
K-Jo: sub-filter is nowhere else defined.  related to comment above

PR #154, opened 2019-11-15
Error not equal 1

ashtul:


ashtul:
If error equal 1, log(error) is 0 and a filter with 0 size is being created

PR #180, opened 2020-01-26
Lift UINT32_MAX limitation on bf.reserve

ashtul:


ashtul:
@gkorland test remains as it will fail on the RedisModule_StringToDoublefunction

PR #181, opened 2020-01-28
Raised APBF minimum capacity to 64

CBaquero:
Raised minimum capacity to 64 to ensure that it is always greater than l and avoid a mod by 0 when shifting.

### DIFF
--- a/src/rm_apbf.c
+++ b/src/rm_apbf.c
@@ -39,8 +39,11 @@ static int parseCreateArgs(RedisModuleCtx *ctx, RedisModuleString **argv, int ar
     if ((RedisModule_StringToLongLong(argv[2], error) != REDISMODULE_OK) || *error < 1 || *error > 5) {
         INNER_ERROR("APBF: invalid error");
     }
-    if ((RedisModule_StringToLongLong(argv[3], capacity) != REDISMODULE_OK) || *capacity < 32) {
-        INNER_ERROR("APBF: invalid capacity or less than 32");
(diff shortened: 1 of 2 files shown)


src/rm_apbf.c
```diff
@@ -39,8 +39,11 @@ static int parseCreateArgs(RedisModuleCtx *ctx, RedisModuleString **argv, int ar
     if ((RedisModule_StringToLongLong(argv[2], error) != REDISMODULE_OK) || *error < 1 || *error > 5) {
         INNER_ERROR("APBF: invalid error");
     }
-    if ((RedisModule_StringToLongLong(argv[3], capacity) != REDISMODULE_OK) || *capacity < 32) {
-        INNER_ERROR("APBF: invalid capacity or less than 32");
+    // A minimum capacity of 64 ensures that assert (capacity > l) will hold
+    // for all (k,l) configurations, since max l is 63.
+    // It is possible to allow smaller capacity by finner testing against actual l
+    if ((RedisModule_StringToLongLong(argv[3], capacity) != REDISMODULE_OK) || *capacity < 64) {
```
ashtul: @CBaquero 
Let's change it to 128 or even 256 in case we change for level 5.

CBaquero: Yes, 128 should accommodate level 5. I didn't notice the code still allowed level 5 configurations in some cases. Another safe option is to limit it to level 4 for now and keep 64.

The future TTL variant will tell us if high level configurations are useful.

PR #198, opened 2020-04-19
Move circle to redis 6-rc3

ashtul:


itamarhaber:
@ashtul rc4 was released the other day, why not that?

ashtul:
@itamarhaber rc4 isn't available yet on dockerhub

PR #201, opened 2020-05-19
Update README.md

banker:
minor README copy fixes


gkorland:
@banker why did you close this PR? it seems like we should fix the README

banker:
@gkorland I'm going to submit a separate PR today.

PR #202, opened 2020-05-20
fix failed PR from forks

ashtul:
based on graph [PR #864](https://github.com/RedisGraph/RedisGraph/commit/0b997c69e2957d9bf7740cc370c7bb41313afbe2#diff-1d37e48f9ceff6d8030570cd36286a61)


gkorland:
@ashtul I think you should move `name: Coverage` above the fork stop

PR #203, opened 2020-05-20
fix Pyhton 3 reads ASCII env params

ashtul:


.circleci/config.yml
```diff
@@ -71,13 +73,15 @@ jobs:
       - run:
           name: Package
           command: >-
```
rafie, line 75: `>-` should be replaced with `|`
and the ramp command "collected" into a single line.

.circleci/config.yml
```diff
@@ -50,10 +50,15 @@ jobs:
     steps:
       - attach_workspace:
           at: /workspace
+      - run:
+          name: Set Environment Variables
```
rafie, line 54: This is local to the shell that sets the variables, and doesn't "reach" the ramp commands. So, if the ramp works, you can remove these entirely.

.circleci/config.yml
```diff
@@ -70,14 +70,14 @@ jobs:
...
-            /workspace/$MODULE_ARTIFACT
+          command: |
+            export LC_ALL=C.UTF-8
+            export LANG=C.UTF-8
+            python -m RAMP.ramp pack -m /workspace/ramp.yml -o /workspace/build/$PACKAGE_NAME.{os}-{architecture}.{semantic_version}.zip /workspace/$MODULE_ARTIFACT
       - run:
           name: Package
           command: >-
-            ramp pack -m /workspace/ramp.yml -o
+            python -m RAMP.ramp pack -m /workspace/ramp.yml -o
```
rafie, line 80: #metoo

PR #217, opened 2020-07-08
Fix client hangs on BF.INSERT with multiple new items when non scaling filter is full

### LINKED ISSUE #216: BF.INSERT with multiple new items when non scaling filter is full leads to clients hanging indefinetly for array elements that will not be sent
You can easily replicate this with Docker and redis-cli:
Start RedisBloom container:
```
docker run -p 6379:6379 redislabs/rebloom:latest
```
on redis-cli:
```
redis-cli BF.INSERT nonscaling_err CAPACITY 3 NONSCALING ITEMS a b c d e
```

tcp stream request portion
```
*11
$9
BF.INSERT
$14
nonscaling_err
$8
CAPACITY
$1
3
$10
NONSCALING
$5
ITEMS
$1
a
$1
b
$1
c
$1
d
$1
e
```
tcp stream reply ( notice that we're explicitly stating we'll reply 5 array elements and we only reply 4 ( up to the first error ) 
```
*5
:1
:1
:1
-Non scaling filter is full
```

This issue should be fixed at `bfInsertCommon` for which I will send PR right away

filipecosta90:
fixes #216

### DIFF
--- a/src/rebloom.c
+++ b/src/rebloom.c
@@ -218,7 +218,8 @@ static int bfInsertCommon(RedisModuleCtx *ctx, RedisModuleString *keystr, RedisM
                           size_t nitems, const BFInsertOptions *options) {
     RedisModuleKey *key = RedisModule_OpenKey(ctx, keystr, REDISMODULE_READ | REDISMODULE_WRITE);
     SBChain *sb;
-    int status = bfGetChain(key, &sb);
+    const int status = bfGetChain(key, &sb);
+    int result = REDISMODULE_OK;
     if (status == SB_EMPTY && options->autocreate) {
         sb = bfCreateChain(key, options->error_rate, options->capacity, options->expansion, options->nonScaling);
         if (sb == NULL) {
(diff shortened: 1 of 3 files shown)


src/rebloom.c
```diff
@@ -229,20 +230,28 @@ static int bfInsertCommon(RedisModuleCtx *ctx, RedisModuleString *keystr, RedisM
...
-    for (size_t ii = 0; ii < nitems; ++ii) {
+    size_t array_len = 0;
+    for (size_t ii = 0; ii < nitems && result == REDISMODULE_OK; ++ii) {
         size_t n;
         const char *s = RedisModule_StringPtrLen(items[ii], &n);
         int rv = SBChain_Add(sb, s, n);
         if (rv == -2) { // decide if to make into an error
-            return RedisModule_ReplyWithError(ctx, "Non scaling filter is full");
+            RedisModule_ReplyWithError(ctx, "Non scaling filter is full");
+            result = REDISMODULE_ERR;
```
ashtul, line 243: what are you doing with the error reply?
I think we should still return OK and the user will see the error in the array.

tests/pytests.py
```diff
@@ -297,6 +297,13 @@ def test_no_scaling(self):
         self.assertOk(self.cmd('bf.reserve bfscale 0.001 1000'))
         self.assertLess(self.cmd('bf.info bfnonscale')[3], self.cmd('bf.info bfscale')[3])
 
+    def test_nonscaling_err(self):
+        capacity = 3
+        self.assertEqual(self.cmd('BF.INSERT nonscaling_err CAPACITY 3 NONSCALING ITEMS a b c'), [1L,1L,1L])
+        # clients hangs
+        self.cmd('BF.INSERT nonscaling_err ITEMS a b c d d')
```
ashtul, line 304: please test for reply string

src/rebloom.c
filipecosta90: @ashtul to keep the same behaviour as before, we need to return `REDISMODULE_ERR`

tests/pytests.py
filipecosta90: agree. fixed in the new commit

src/rebloom.c
ashtul: `RedisModule_ReplyWithError` always returns `REDISMODULE_OK`

filipecosta90: you're totally right. fixed on the new commit

Issue #115, opened 2019-09-05
BloomFilter - Expiration on items


K-Jo:

asad-awadia:
Any updates on this? @K-Jo

ashtul:
@asad-awadia We currently develop an algorithm that will support such behavior.
It is work in progress and will take some time to be included.
I will update you as soon as I have some news.

CBaquero:
There is now an experimental branch "AgePartitionedBF" with a APBF bloom filter that stores a configurable number of the most recently added elements, older elements will gradually expire.
Details in https://arxiv.org/abs/2001.03147

DeltaLaboratory:
Any updates on this feature?

asad-awadia:
wow been 4 years - time flies

PR #238, opened 2020-10-06
Time-based APBF

Nimbus79:
related to #115

### DIFF
--- a/contrib/agingBloom.c
+++ b/contrib/agingBloom.c
@@ -14,7 +14,6 @@
 #include <stdlib.h>     // calloc
 #include <string.h>     // memset
 #include <unistd.h>     // write
-#include <sys/time.h>   // gettimeofday
 
 #include "hash.h"
 #include "agingBloom.h"
@@ -22,8 +21,6 @@
 #define _64BITS 64
 #define BYTE 8
 
-struct timeval start, end;  // Used to record time between insertions.
-timestamp_t ts;             // Timestamp in seconds.
 
 /******************* Static functions ***************/
 static void SetBitOn(char *array, uint64_t loc) {
@@ -92,7 +89,7 @@ static void destroySlice(blmSlice *slice) {
     slice->data = NULL;
 }
 
-static uint64_t slice_capacity(blmSlice *slice) {
+static uint64_t sliceCapacity(blmSlice *slice) {
     return log(2) * slice->size;
 }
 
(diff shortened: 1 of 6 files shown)


ashtul:
@CBaquero @pssalmeida 
You can add leave here nay comments you may have.

contrib/agingBloom.c
```diff
@@ -94,315 +98,410 @@ static void destroySlice(blmSlice *slice) {
...
+        apbf->numSlices = ++i;
+        slices = (blmSlice *) realloc(slices, sizeof(blmSlice) * i);
+        assert(slices);
+        apbf->slices = slices;
     }
-    break;
-  }
 }
 
+// Calculates the new size of slice 0.
```
ashtul, line 172: add explanation?

ashtul (commented):
Looks great.
I've pushed some minor changes including a suggestion by @pssalmeida to avoid looping over `memcpy`.

Two additional small comments.

ashtul (changes requested):
We get a warning from [LGTM](https://lgtm.com/projects/g/RedisBloom/RedisBloom/rev/pr-62bbec16678d5d011841f6c1d28071f6049b971c). Why are these global variables?

We should probably have `start`, `end` and `ts` as part of the struct since we could have multiple filters.

CBaquero:
Yes, these variables for time tracking look like they should be local to each filter.

contrib/agingBloom.h
```diff
@@ -28,59 +28,67 @@
...
 #define CHECK_ALL 0
 
+//#define SLEEP 0.1     // Used only when static time is wanted between insertions
+                        // instead of real time. (defined in seconds)
+
 #define TWOHASH
 
 typedef struct ageBloom_s ageBloom_t;
-typedef uint64_t timestamp_t;
+typedef double_t timestamp_t;
```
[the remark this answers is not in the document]
CBaquero: Can you check, @Nimbus79, if they can stay as uint64_t?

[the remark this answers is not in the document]
Nimbus79: I switched it to double so I could save the timestamp in seconds since I obtain it in microseconds. I assumed the user-specified time span was in seconds as well, so that's why I converted everything to seconds.

contrib/agingBloom.c
```diff
@@ -94,315 +102,446 @@ static void destroySlice(blmSlice *slice) {
...
-  int8_t level = 4;
-  ageBloom_t *apbf = APBF_createHighLevelAPI(error, capacity, level);
-  apbf->assessFreq = frequency;
-  return apbf;
+// Used for time based APBF
+ageBloom_t *APBF_createTimeAPI(int error, uint64_t capacity, int8_t level, timestamp_t frequency) {
+    ageBloom_t *apbf = APBF_createHighLevelAPI(error, capacity, level);
+    assert(apbf);
+    ts = apbf->assessFreq = frequency;
+    apbf->slices[apbf->numHash].timestamp = ts;
```
ashtul, line 407: What are we doing here?

Nimbus79: I changed some of the names as you suggested in our last meeting: ts is now currTime and assessFreq is timeSpan, but we can change them again if they're not good.
Initially, I set the currTime to be equal to the timeSpan, so when we're doing the verification whether a slice is in the specified time span (slices[i].timestamp < **apbf->currTime - apbf->timeSpan**) we don't get a negative value. The same goes for the second line of code when the time between shifts is calculated in the function predictSize.
In alternative, I could also set the currTime initially to 0 and, every time the verification is made, simply add to it the value of timeSpan, like this: slices[i].timestamp < (apbf->currTime + timeSpan) - timeSpan. Maybe the second option is better because the variable currTime doesn't grow so fast, since it starts as 0 instead of a large number. But there's also the downside of having to do this add operation in more than one place every time there's a shift. What do you think?

PR #248, opened 2020-10-23
Add CMS.DELETE

jrideout:
Deletes the key


jrideout:
So I missed something obvious: `del` works just fine as far as I can tell, so I'm not sure this patch is needed. That said, if `del` is the supported operation, then the docs should be updated to reflect that.

ashtul:
@jrideout You are correct. `DEL` will remove any key from the keyspace.
Thank you for the effort.

PR #258, opened 2020-12-06
fix BFAofRewrite() save header`s iter not equal 1

kukey:
when the key not exist, the loadchunk read header will fail, because the iter not equal 1.


kukey:
or the redis-server restart, load data from aof, the command will load fail.

ashtul (approved):
Great bugfix!
Thank you

ashtul:
@kukey Please sign the CLA so your fix can be merged.

Issue #259, opened 2020-12-07
cuckoo filter initialization parameters


qingping209:
testBulkDel of test-cuckoo.c will fail if expansion was set to 2 instead of 1.  why was that? 

![image](https://user-images.githubusercontent.com/11344292/101300462-a32c1580-3870-11eb-8671-a2681fe44010.png)

![image](https://user-images.githubusercontent.com/11344292/101300475-af17d780-3870-11eb-8d00-63e19b8a5b1b.png)

does the correctness depend on how the filter is initialized ?

qingping209:
@ashtul

ashtul:
@qingping209 , thank you for the issue.
This seems odd. I'll take a look.

qingping209:
> @qingping209 , thank you for the issue.
> This seems odd. I'll take a look.

and it seems like *expansion* is the only the variable that would fail the test. and also I noticed in rebloom.c the default value of expansion is also 1.   

![image](https://user-images.githubusercontent.com/11344292/101320415-af2ecc00-389e-11eb-8b41-94053fb7e7c6.png)

qingping209:
any update on this ? @ashtul

qingping209:
fixed by #260

PR #260, opened 2020-12-07
Fix Cuckoo filter compact cause deleted by mistake

jingjunLi:
This PR fixes the issue reported on [#259](https://github.com/RedisBloom/RedisBloom/issues/259).

This is can be reproduced using the test-cuckoo.c file (testBulkDelwithExpansion function) without change. Because the Cuckoo Filter Compact only relocate previous `cf->numBuckets`, but not the total SubCF.

```sh
--- TEST cuckoo.testBulkDelwithExpansion (6/7) ---
Assertion failed at test-cuckoo.c:185
Expected: 1 == CuckooFilter_Delete(&ck, MurmurHash64A_Bloom(&ii, sizeof ii, 0))
        1: U=1 I=1 H=0x1
        CuckooFilter_Delete(&ck, MurmurHash64A_Bloom(&ii, sizeof ii, 0)): U=0 I=0 H=0x0
Aborted (core dumped)
```

### DIFF
--- a/src/cuckoo.c
+++ b/src/cuckoo.c
@@ -180,8 +180,8 @@ uint64_t CuckooFilter_Count(const CuckooFilter *filter, CuckooHash hash) {
 int CuckooFilter_Delete(CuckooFilter *filter, CuckooHash hash) {
     LookupParams params;
     getLookupParams(hash, &params);
-    for (uint16_t ii = 0; ii < filter->numFilters; ++ii) {
-        if (Filter_Delete(&filter->filters[ii], &params)) {
+    for (uint16_t ii = filter->numFilters; ii > 0; --ii) {
+        if (Filter_Delete(&filter->filters[ii - 1], &params)) {
             filter->numItems--;
             filter->numDeletes++;
             if (filter->numFilters > 1 && filter->numDeletes > (double)filter->numItems * 0.10) {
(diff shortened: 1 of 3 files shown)


src/cuckoo.c
```diff
@@ -180,8 +180,8 @@ uint64_t CuckooFilter_Count(const CuckooFilter *filter, CuckooHash hash) {
 int CuckooFilter_Delete(CuckooFilter *filter, CuckooHash hash) {
     LookupParams params;
     getLookupParams(hash, &params);
-    for (uint16_t ii = 0; ii < filter->numFilters; ++ii) {
-        if (Filter_Delete(&filter->filters[ii], &params)) {
+    for (uint16_t ii = filter->numFilters; ii > 0; --ii) {
+        if (Filter_Delete(&filter->filters[ii - 1], &params)) {
```
ashtul, line 184: Why make this change?
The idea here is to clear the latest filter which is likely to be the one with the least items in it.

src/cuckoo.c
```diff
@@ -339,14 +339,15 @@ static int relocateSlot(CuckooFilter *cf, CuckooBucket bucket, uint16_t filterIx
...
+    MyCuckooBucket *filter = currentFilter->data;
     int dirty = 0;
     uint64_t numRelocs = 0;
 
-    for (uint64_t bucketIx = 0; bucketIx < cf->numBuckets; ++bucketIx) {
-        for (uint16_t slotIx = 0; slotIx < cf->bucketSize; ++slotIx) {
-            int status =
-                relocateSlot(cf, &filter[bucketIx * cf->bucketSize], filterIx, bucketIx, slotIx);
+    for (uint64_t bucketIx = 0; bucketIx < currentFilter->numBuckets; ++bucketIx) {
+        for (uint16_t slotIx = 0; slotIx < currentFilter->bucketSize; ++slotIx) {
```
ashtul, line 348: This is where the issue resides.

ashtul (commented):
@jingjunLi 
Great PR. Thank you!
One comment.

src/cuckoo.c
ashtul: My suspicion is this change just hides the real issue here.

qingping209: > My suspicion is this change just hides the real issue here.

In the original code, insertion is bottom up but deletion is top down. 
This miss match can cause the following anomaly:

suppose we insert the following keys a c and
suppose fingerprint(a) == fingerprint(c) and
hash(c) % filter0.numBucket == hash(a) % filter0.numBuckets
hash(c) % filter1.numBucket != hash(a) % filter1.numBuckets

a resided in level0 filter
c reside in level1 filter

filter0: a 
fitler1: c

now we try to delete c from the filter, and since deletion is performed top down; so 'a' is deleted *by mistake*.  
then we try to delete a from the filter, now we cannot find 'a' in fitler1 (it's already deleted from fitler0);
so as long the an expansion of rate > 1 happens and the insertion and deletion order mismatch, deletion may fail.

ashtul: @qingping209 
You are right! We have to delete hashes from the larger to the smaller. It was not apparent before since sizes for all filters were identical.
Thank you very much for this contribution.

ashtul (approved):

ashtul:
@jingjunLi Can you please sign the CLA so I can merge this PR?

jingjunLi:
> @jingjunLi Can you please sign the CLA so I can merge this PR?

jingjunLi:
> @jingjunLi Can you please sign the CLA so I can merge this PR?

I haved signed the CLA.  Thanks !

ashtul (approved):

PR #286, opened 2021-04-06
Create codeql-analysis.yml

gkorland:


ashtul (approved):

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,67 @@
...
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ master ]
```
chayim, line 19: Recommend you add versioned branches        - '[0-9]+.[0-9]+'

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,67 @@
...
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ master ]
```
chayim, line 16: Recommend you add versioned branches        - '[0-9]+.[0-9]+'

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,67 @@
...
+  pull_request:
```
chayim, line 17: We can remove the pull request action check if we maintain an on-push to versioned branches

chayim (changes requested):
See comments re versioned branches.

.github/workflows/codeql-analysis.yml
gkorland: with pull request you can check the code before you merge it

.github/workflows/codeql-analysis.yml
gkorland: I tried it, but it seems like they don't know how to parse it...

chayim (approved):

PR #299, opened 2021-05-02
Using normalized Dockerfile ( with python3 ) and system-setup.py

filipecosta90:
Following @chayim pinpoint about docker steps failing due to the current changes not working on python2 env I believe the best approach is to normalize the used dockerfile and system-setup.py so we have the same look and feel across all steps.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,7 +18,8 @@ commands:
     steps:
       - run:
           name: Pull Submodules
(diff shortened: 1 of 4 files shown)


rafie:
@filipecosta90 It's a pity we're not synchronized and doing double work. I should have pushed my PR as draft.

rafie (approved):

filipecosta90:
closing in favour of #300

PR #306, opened 2021-05-08
[bug] fix bf.loadchunk replicate failed

luoming1224:
when BFLoadChunk_RedisCommand function is called for the first time, in other words, parmeter iter == 1,  RedisModule_ReplicateVerbatim function was not called after success. Causes the loadchunk command not to replicate to the replica node.

### DIFF
--- a/src/rebloom.c
+++ b/src/rebloom.c
@@ -469,6 +469,7 @@ static int BFLoadChunk_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **arg
             return RedisModule_ReplyWithError(ctx, errmsg);
         } else {
             RedisModule_ModuleTypeSetValue(key, BFType, sb);
+            RedisModule_ReplicateVerbatim(ctx);
             return RedisModule_ReplyWithSimpleString(ctx, "OK");
         }
     } else if (status != SB_OK) {


ashtul:
@luoming1 thank you for the contribution. :+1: 
Please sign the CLA so your contribution can be merged.

ashtul:
@luoming1224, are you able to reproduce an error resulted from the lack of `RedisModule_ReplicateVerbatim`?
You are completely right about the fix, but I wonder how did you notice it.
Thank you

luoming1224:
> the lack of

Here is part of the code in my test case
`List<ChunkData> chunkDataList = new ArrayList<>();
        long iter = 0;
        while (true) {
            ChunkData chunkData = bloomFilter.scanDump(key, iter);
            iter = chunkData.getIter();
            if (0 == iter) {
                break;
            }

            chunkDataList.add(chunkData);
        }

        jedis.del(key);

        String res = null;
        for (ChunkData chunkData : chunkDataList){
            res = bloomFilter.loadChunk(key, chunkData);
            Assert.assertEquals(res, "OK");
        }
        info = bloomFilter.info(key);`

I used **read-write separation** when testing the Bloom filter in redis, and when verifying the loadChunk command, we used the BF.INFO command to further verify the correctness of the loadChunk command after it was executed, so we found that the loadChunk command was executed correctly and return" OK", but the BF.INFO command returns an "ERR not found" error.
So I used redis-cli to verify the problem, and it was consistent with expectations. The master node data is normal, but the replica node does not have the key.
Confirm that loadchunk is a write command in the code, no problem
Using GDB to debug in redis, it is found that server.dirty=0 when the loadChunk command is executed for the first time.
Review the code and find that the RedisModule_ReplicateVerbatim function was not executed when the loadchunk command was successfully executed for the first time.

ashtul:
@luoming1224 Sure, Please let's try one more time.

ashtul:
Replaced by #309

PR #321, opened 2021-06-11
Kick off CI benchmarks automation. Included BF.ADD benchmark

filipecosta90:
The following PR introduces the CI benchmark automation. 
The automated benchmark definitions provides a framework for evaluating and comparing feature branches and catching regressions prior to letting them into the master branch.
The full description of how we handle can be seen at: https://github.com/RedisLabsModules/redisbench-admin/tree/master/docs and within the added `tests/benchmarks/Readme.md` you will find more details.


.circleci/config.yml
```diff
@@ -72,6 +72,45 @@ commands:
...
+      module_path:
+        type: string
+        default: ../../redisbloom.so
+    steps:
+      - run:
+          name: Install remote benchmark tool dependencies
+          command: |
+            VERSION=0.14.8 ./deps/readies/bin/getterraform
+      - run:
+          name: Install remote benchmark python dependencies
```
chayim, line 89: WDY of embracing poetry like in the modules?

chayim (approved):
Approved - but I suggest embracing poetry.

.circleci/config.yml
filipecosta90: If we see we want to include it on all modules then 100% agree with it :+1:

ashtul (approved):

PR #328, opened 2021-07-13
Using redis allocator within T-Digest sketch

### LINKED ISSUE #327: T-digest structure should use redis default allocator
The data structure is prepared to do so, via 

We just need to define `TD_ALLOC_H` prior including the `tdigest.h`, 

as explained bellow ( and in https://github.com/RedisBloom/t-digest-c/blob/master/src/td_malloc.h ) :

```
 * This file is used in order to change the t-digest allocator at compile time.
 * Just define the following defines to what you want to use. Also add
 * the include of your alternate allocator if needed (not needed in order
 * to use the default libc allocator). */

#ifndef TD_ALLOC_H
#define TD_ALLOC_H
#define __td_malloc malloc
#define __td_calloc calloc
#define __td_realloc realloc
#define __td_free free
#endif
```

filipecosta90:
Fixes #327

### DIFF
--- a/src/rm_tdigest.c
+++ b/src/rm_tdigest.c
@@ -2,12 +2,20 @@
 #include <stdlib.h>  // malloc
 #include <strings.h> // strncasecmp
 
-#include "tdigest.h"
+#include "rm_tdigest.h"
 #include "rmutil/util.h"
 #include "version.h"
-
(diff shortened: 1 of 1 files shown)


ashtul (approved):
LGTM 👍

PR #331, opened 2021-07-20
add WITHCOUNT flag to TOPK.LIST 

### LINKED ISSUE #324: Need a command to return all TOP-K entries with their counts
Now I could do this using 2 queries(TOPK.LIST + TOPK.COUNT) but while doing so some items could have been expelled which is non ideal...

ashtul:
fixes #324

### DIFF
--- a/docs/TopK_Commands.md
+++ b/docs/TopK_Commands.md
@@ -184,12 +184,13 @@ TOPK.COUNT test foo 42 nonexist
 Return full list of items in Top K list.
 
 ```
-TOPK.LIST {key}
+TOPK.LIST {key} [WITHCOUNT]
 ```
 
 ### Parameters
 
 * **key**: Name of sketch where item is counted.
+* **WITHCOUNT**: Count of each element is returned.  
 
 ### Complexity
 
--- a/src/rm_topk.c
+++ b/src/rm_topk.c
@@ -214,7 +214,7 @@ static int TopK_List_Cmd(RedisModuleCtx *ctx, RedisModuleString **argv, int argc
             if (withCount) {
                 RedisModule_ReplyWithLongLong(ctx, heapList[i].count);
             }
-            arrlen += withCount ? 2 : 1;
+            arrlen += 1 + withCount;
         } else {
             break;
         }
--- a/src/rm_topk.c
+++ b/src/rm_topk.c
@@ -187,25 +187,40 @@ static int TopK_Count_Cmd(RedisModuleCtx *ctx, RedisModuleString **argv, int arg
 }
 
 static int TopK_List_Cmd(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
-
-    if (argc != 2)
+    if (argc != 2 && argc != 3)
         return RedisModule_WrongArity(ctx);
+
+    bool withCount = 0;
+    if (argc == 3) {
+        size_t len;
+        const char *countStr = RedisModule_StringPtrLen(argv[2], &len);
+        if (strncasecmp(countStr, "WITHCOUNT", len) == 0) {
+            withCount = 1;
+        } else {
+            return RedisModule_ReplyWithError(ctx, "WITHCOUNT keyword expected");
+        }
+    }
 
     TopK *topk = NULL;
     if (GetTopKKey(ctx, argv[1], &topk, REDISMODULE_READ) != REDISMODULE_OK) {
         return REDISMODULE_OK;
     }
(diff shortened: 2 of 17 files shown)


oshadmi (commented):
Few questions

src/rm_topk.c
```diff
@@ -187,25 +187,40 @@ static int TopK_Count_Cmd(RedisModuleCtx *ctx, RedisModuleString **argv, int arg
...
     }
-    uint32_t k = topk->k;
-    char **heapList = TOPK_CALLOC(k, (sizeof(char *)));
-    TopK_List(topk, heapList);
-    RedisModule_ReplyWithArray(ctx, k);
-    for (int i = 0; i < k; ++i) {
-        if (heapList[i] != NULL) {
-            RedisModule_ReplyWithSimpleString(ctx, heapList[i]);
+    HeapBucket *heapList = TopK_List(topk);
+    RedisModule_ReplyWithArray(ctx, REDISMODULE_POSTPONED_ARRAY_LEN);
```
filipecosta90, line 209: @ashtul given we know there are implications on `REDISMODULE_POSTPONED_ARRAY_LEN` and that we know beforehand the size of the array ( either k or 2xk ) lets keep RedisModule_ReplyWithArray with the proper arraylen. 
agree?

docs/TopK_Commands.md
```diff
@@ -184,12 +184,13 @@ TOPK.COUNT test foo 42 nonexist
...
-TOPK.LIST {key}
+TOPK.LIST {key} [WITHCOUNT]
 ```
 
 ### Parameters
 
 * **key**: Name of sketch where item is counted.
+* **WITHCOUNT**: Count of each element is returned.  
 
 ### Complexity
```
filipecosta90, line 196: @ashtul we're missing the changes to the reply. 
Also, we need to propagate the awareness of this change into all clients.

filipecosta90:
@ashtul this a good opportunity to add a topk benchmark variation. 
Added some other comments about changes that IMHO can be avoided, and some missing docs.

src/rm_topk.c
ashtul: We might break with less than `k` elements if the heap is not full yet.

src/rm_topk.c
```diff
@@ -187,25 +187,40 @@ static int TopK_Count_Cmd(RedisModuleCtx *ctx, RedisModuleString **argv, int arg
...
+    long arrlen = 0;
+    for (int i = 0; i < topk->k; ++i) {
+        if (heapList[i].count != 0) {
```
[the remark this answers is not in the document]
ashtul: Before we had `char **` so `heapList[i]` held a pointer. Now we hold a struct that cannot be `NULL`.
This is why `cmpHeapBucket` is safe.

docs/TopK_Commands.md
ashtul: We don't always put output for all optional params as it makes the docs very long.

ashtul (commented):
Comments answered or fixed.

src/rm_topk.c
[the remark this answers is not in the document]
oshadmi: If command is as before, without `WITHCOUNT`, would sorting and changing the order be a breaking change?
We could keep the previous behavior in this case (use pointers, do not sort, and avoid the REDISMODULE_POSTPONED_ARRAY_LEN) and also have better performance,
and only have the new behavior with `WITHCOUNT`.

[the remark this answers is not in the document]
ashtul: @oshadmi Changing the order should not be considered breaking change IMO.

src/rm_topk.c
ashtul: @filipecosta90 a single use of `REDISMODULE_POSTPONED_ARRAY_LEN` would not hurt performance. From what I know, we had an issue in search where we had wide use of it in a reply true of possibly millions of results.
The other way to do it is to iterate twice over the array and count the legal result but this seems excessive to me. WDYT?

filipecosta90:
@ashtul can we take this opportunity to add a benchmark definition for TOPK.LIST? 
Like we have for bf.add https://github.com/RedisBloom/RedisBloom/blob/master/tests/benchmarks/bf_add_cap10M_err0.001.yml

oshadmi (approved):
LGTM 👍🏼

filipecosta90 (approved):
LGTM. Thank you for the extra effort put into adding more benchmarks to this!

PR #350, opened 2021-08-26
MOD: (CMS) allow batch key and value

JinxLbj:
1. (CMS) allow batch key and value;

### DIFF
--- a/src/rm_cms.c
+++ b/src/rm_cms.c
@@ -184,15 +184,27 @@ int CMSketch_Query(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
 
 int CMSketch_BatchQuery(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
     RedisModule_AutoMemory(ctx);
+    int ignore_key = 0;
     if (argc < 5) {
         return RedisModule_WrongArity(ctx);
     }
     int pos1 = RMUtil_ArgIndex("KEYS", argv, argc);
     int pos2 = RMUtil_ArgIndex("VALUES", argv, argc);
+    int nx = RMUtil_ArgIndex("NX", argv, argc);
+    int xx = RMUtil_ArgIndex("XX", argv, argc);
     int keyCount = pos2 - pos1 - 1;
     int itemCount = argc - 3 - keyCount;
-    if (pos1 == -1 || pos2 == -1 || keyCount <= 0 || itemCount <= 0) {
-        INNER_ERROR("CMS: BATCHQUERY MUST BE: cms.batchquery KEYS k1 k2 k3... VALUES v1 v2 v3...");
+    if (nx == argc - 1) {
+        ignore_key = 1;
+        itemCount--;
+    } else if (xx == argc - 1) {
+        ignore_key = 0;
+        itemCount--;
+    }
+    if (pos1 == -1 || pos2 == -1 || keyCount <= 0 || itemCount <= 0 ||
+        (nx != -1 && nx != argc - 1) || (xx != -1 && xx != argc - 1)) {
+        INNER_ERROR(
+            "CMS: BATCHQUERY MUST BE: cms.batchquery KEYS k1 k2 k3... VALUES v1 v2 v3... [XX/NX]");
     }
     unsigned int counter[itemCount];
     memset(counter, 0, sizeof(counter));
@@ -201,7 +213,10 @@ int CMSketch_BatchQuery(RedisModuleCtx *ctx, RedisModuleString **argv, int argc)
     int depth = 0;
     for (int i = pos1 + 1; i < pos2; ++i) {
         CMSketch *cms = NULL;
-        if (GetCMSKeyIgnore(ctx, argv[i], &cms, REDISMODULE_READ) == REDISMODULE_OK) {
+        if ((ignore_key == 1 &&
+             GetCMSKeyIgnore(ctx, argv[i], &cms, REDISMODULE_READ) == REDISMODULE_OK) ||
+            (ignore_key == 0 &&
+             GetCMSKey(ctx, argv[i], &cms, REDISMODULE_READ) == REDISMODULE_OK)) {
             if (depth != 0 && depth != cms->depth) {
                 INNER_ERROR("CMS: all keys must have same depth");
             } else {
--- a/src/rm_cms.c
+++ b/src/rm_cms.c
@@ -7,6 +7,9 @@
 
 #include "cms.h"
 #include "rm_cms.h"
+#include "contrib/murmurhash2.h"
+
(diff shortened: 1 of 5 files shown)


src/rm_cms.c
```diff
@@ -166,6 +182,62 @@ int CMSketch_Query(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
...
+        return RedisModule_WrongArity(ctx);
+    }
+    int pos1 = RMUtil_ArgIndex("KEYS", argv, argc);
+    int pos2 = RMUtil_ArgIndex("VALUES", argv, argc);
+    int keyCount = pos2 - pos1 - 1;
+    int itemCount = argc - 3 - keyCount;
+    if (pos1 == -1 || pos2 == -1 || keyCount <= 0 || itemCount <= 0) {
+        INNER_ERROR("CMS: BATCHQUERY MUST BE: cms.batchquery KEYS k1 k2 k3... VALUES v1 v2 v3...");
+    }
+    unsigned int temp[itemCount];
```
ashtul, line 197: Let's find a descriptive name. Maybe `totalCount`?

src/rm_cms.c
```diff
@@ -166,6 +182,62 @@ int CMSketch_Query(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
...
+            } else {
+                depth = cms->depth;
+            }
+            key_arr[key_index++] = cms;
+        }
+    }
+    RedisModule_ReplyWithArray(ctx, itemCount);
+    // set 0
+    memset(temp, 0, sizeof(temp));
+    // if all keys don't exists, return all values 0
```
ashtul, line 215: Clean and move near decleration.

tests/flow/test_cms.py
```diff
@@ -107,6 +114,30 @@ def test_incrby_query(self):
...
+        self.assertOk(self.cmd('cms.initbydim', 'k2', '100', '5'))
+        self.assertOk(self.cmd('cms.initbydim', 'k3', '100', '6'))
+        self.assertEqual([5], self.cmd('cms.incrby', 'k1', 'v1', '5'))
+        self.assertEqual([5], self.cmd('cms.incrby', 'k2', 'v2', '5'))
+        self.assertEqual([5], self.cmd('cms.incrby', 'k3', 'v3', '5'))
+        self.assertEqual([5, 5], self.cmd('cms.batchquery', 'KEYS', 'k1', 'k2', 'VALUES', 'v1', 'v2'))
+        # k1, k3 have diff depth, error
+        self.assertRaises(ResponseError, self.cmd, 'cms.batchquery', 'KEYS', 'k1', 'k3', 'VALUES', 'v1', 'v3')
+        # at least one key existed, k1 existed
+        self.assertEqual([5, 0], self.cmd('cms.batchquery', 'KEYS', 'k1', 'k4', 'VALUES', 'v1', 'v4'))
```
ashtul, line 128: This should return an error as well.

tests/flow/test_cms.py
```diff
@@ -107,6 +114,30 @@ def test_incrby_query(self):
...
         #     self.assertEqual([0], self.cmd('cms.query', 'test', 'nonexist'))
 
+    def test_batchquery(self):
+        self.assertOk(self.cmd('cms.initbydim', 'k1', '100', '5'))
+        self.assertOk(self.cmd('cms.initbydim', 'k2', '100', '5'))
+        self.assertOk(self.cmd('cms.initbydim', 'k3', '100', '6'))
+        self.assertEqual([5], self.cmd('cms.incrby', 'k1', 'v1', '5'))
+        self.assertEqual([5], self.cmd('cms.incrby', 'k2', 'v2', '5'))
+        self.assertEqual([5], self.cmd('cms.incrby', 'k3', 'v3', '5'))
+        self.assertEqual([5, 5], self.cmd('cms.batchquery', 'KEYS', 'k1', 'k2', 'VALUES', 'v1', 'v2'))
```
ashtul, line 124: Can we plese have an example with a count from two keys?

src/rm_cms.c
```diff
@@ -32,6 +35,19 @@ static int GetCMSKey(RedisModuleCtx *ctx, RedisModuleString *keyName, CMSketch *
     return REDISMODULE_OK;
 }
 
+static int GetCMSKeyIgnore(RedisModuleCtx *ctx, RedisModuleString *keyName, CMSketch **cms,
```
ashtul, line 38: I believe this was added so in case a key is missing, the command can still succeed.
IMO, it should fail.
@filipecosta90 @gkorland WDYT?

ashtul (commented):
@JinxLbj Thank you for an elegant PR. 

There is the issue of whether a command should succeed or fail in case a key is missing. Otherwise, a few minor style and test comments.

I would like to ask you to document this new command at docs/CountMinSketch_Commands.md

src/rm_cms.c
JinxLbj: Thx for review my code.
IMO, The entire request should not be erroneous because one key does not exist in the batch command.
But it's really easy to ignore some errors.
If everyone thinks that exceptions should be thrown, I will fix it.

src/rm_cms.c
JinxLbj: Sorry, i dont' get it.
IMO,
keyCount is the number of the key.
itemCount is the number of the item.
And each of these can't be empty.

tests/flow/test_cms.py
JinxLbj: `self.assertEqual([5, 5], self.cmd('cms.batchquery', 'KEYS', 'k1', 'k2', 'VALUES', 'v1', 'v2'))`
I think it is want you want.

src/rm_cms.c
JinxLbj: Already fixed.

tests/flow/test_cms.py
ashtul: @JinxLbj Let's add  
```
self.assertEqual([5], self.cmd('cms.incrby', 'k1', 'v2', '5'))
```
so the batch query will return [5, 10]

src/rm_cms.c
ashtul: Thank you. I will let you know...

src/rm_cms.c
ashtul: My (unclear) comment was about the use of the name `temp`.

src/rm_cms.c
JinxLbj: Hello, I have modified the code.
Use the [NX/XX] to let the user decide whether to ignore exceptions.
What do you think.

ashtul:
@JinxLbj After further consultation, we won't be able to accept this PR since it introduces Redis command on multiple keys since it cannot run on a cluster.
Thank you for the effort.

Did you consider pipelining the commands to improve performance?

PR #353, opened 2021-09-14
Remove use of implementation detail __compar_fn_t

TedLyngmo:
`TopK_List` uses `__compar_fn_t` to cast `cmpHeapBucket` to the correct
signature and fails to compile using implementations without the
`__compar_fn_t` `typedef`. This patch makes that cast unnecessary.

Signed-off-by: Ted Lyngmo <ted@lyncon.se>


ashtul (approved):
Thank you @TedLyngmo for this fix

PR #385, opened 2021-12-13
PUBLISH_OSNICK, and adding circle variables

chayim:


build/docker/Makefile
```diff
@@ -2,7 +2,7 @@ REDIS_VERSION=6.2.5
 PRODUCT=rebloom
```
rafie, line 2: We can add
```
ROOT=../..
include $(ROOT)/deps/readies/mk/main
```
at the top.

build/docker/Makefile
```diff
@@ -2,7 +2,7 @@ REDIS_VERSION=6.2.5
...
 
 OSNICK ?= buster
-OSNICK_OFFICIAL ?= buster
+PUBLISH_OSNICK ?= buster
```
rafie, line 5: `buster` -> `bullseye`

.circleci/config.yml
```diff
@@ -145,13 +145,13 @@ commands:
       - run:
           name: Build for platform (docker build)
           command: |
-            make -C build/docker build DOCKER_SUFFIX=".<<parameters.platform>>" OSNICK=<<parameters.platform>> TEST=1 VERBOSE=1
+            make -C build/docker build DOCKER_SUFFIX=".<<parameters.platform>>" OSNICK=<<parameters.platform>> CIRCLE_BRANCH=${CIRCLE_BRANCH} CIRCLE_TAG=${CIRCLE_TAG} TEST=1 VERBOSE=1
```
rafie, line 148: Don't we process `CIRCLE_BRANCH`/`CIRCLE_TAG` inside `mk/docker.rules`?
Also, do we still need an explicit `DOCKER_SUFFIX` here?

chayim: Correct - removed in the latest push.

build/docker/Makefile
chayim: Updated in latest

PR #392, opened 2022-01-23
SB_NewChainFromHeader init growth always equal 0

kukey:
1. This may cause redis-server core when the bloom expand.
2. Or use redis comamnd dump, restore it use bf.info command the expansion is 0.


ashtul (approved):
Good catch.

PR #396, opened 2022-02-01
MOD-2082 Add intershard_tls_pass support

gkorland:
RedisBloom support intershard_tls_pass by default since it doesn't have cross shards communication


ashtul (approved):
LGTM
The build fails on linting which is unrelated.

PR #405, opened 2022-02-10
Add rueidis to index.md

rueian:
Hi, I made a Golang Redis client library that supports RedisBloom commands.

Could I add my library to the client list? Thanks!


LiorKogan (changes requested):
The language is often referred to as Golang because of its former domain name, golang.org , but the proper name is Go (see https://tip.golang.org/doc/faq#go_or_golang)

Please change Golang in docs/index.md to Go (two places).

(I'm not sure why we have index.md and docs/index.md, but we'll solve this later).

rueian:
Hi, @LiorKogan 
I have changed them into just Go.

PR #419, opened 2022-02-28
[BUG] give error on CMS overflow

### LINKED ISSUE #418: CMS 32-bit overflow
If a CMS count reaches the max size of a 32-bit int, it will roll over to zero. This violates the never-underestimate-principle of CMS. The value should simply stop incrementing once it reaches the max value of a 32-bit int.

ashtul:
fixes #418 
Return an error message `CMS: INCRBY overflow` if overflow occurs on CMS.INCRBY.
CMS.QUERY will return `UINT32_MAX` or `4294967295`

### DIFF
--- a/src/cms.c
+++ b/src/cms.c
@@ -6,8 +6,7 @@
 #include "cms.h"
 #include "contrib/murmurhash2.h"
 
-#define min(a,b) (((a)<(b))?(a):(b))
-
+#define min(a, b) (((a) < (b)) ? (a) : (b))
 
 #define BIT64 64
 #define CMS_HASH(item, itemlen, i) MurmurHash2(item, itemlen, i)
--- a/src/cms.c
+++ b/src/cms.c
@@ -6,12 +6,8 @@
 #include "cms.h"
 #include "contrib/murmurhash2.h"
 
-#define min(a, b)                                                                                  \
-    ({                                                                                             \
-        __typeof__(a) _a = (a);                                                                    \
-        __typeof__(b) _b = (b);                                                                    \
-        _a < _b ? _a : _b;                                                                         \
-    })
+#define min(a,b) (((a)<(b))?(a):(b))
+
 
 #define BIT64 64
 #define CMS_HASH(item, itemlen, i) MurmurHash2(item, itemlen, i)
@@ -60,12 +56,11 @@ size_t CMS_IncrBy(CMSketch *cms, const char *item, size_t itemlen, size_t value)
         cms->array[loc] += value;
         if (cms->array[loc] < value) {
             cms->array[loc] = UINT32_MAX;
-            overflow = 1;
         }
         minCount = min(minCount, cms->array[loc]);
     }
     cms->counter += value;
-    return !overflow ? minCount : UINT32_MAX;
+    return minCount;
 }
 
 size_t CMS_Query(CMSketch *cms, const char *item, size_t itemlen) {
--- a/src/cms.c
+++ b/src/cms.c
@@ -52,14 +52,20 @@ size_t CMS_IncrBy(CMSketch *cms, const char *item, size_t itemlen, size_t value)
     assert(item);
 
     size_t minCount = (size_t)-1;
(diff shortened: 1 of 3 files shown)


src/cms.c
```diff
@@ -52,14 +52,20 @@ size_t CMS_IncrBy(CMSketch *cms, const char *item, size_t itemlen, size_t value)
...
+        cms->array[loc] += value;
+        if (cms->array[loc] < value) {
+            cms->array[loc] = UINT32_MAX;
+            overflow = 1;
+        }
+        minCount = min(minCount, cms->array[loc]);
     }
     cms->counter += value;
-    return minCount;
+    return !overflow ? minCount : UINT32_MAX;
```
gkorland, line 68:
```suggestion
    return overflow ? UINT32_MAX : minCount;
```

src/rm_cms.c
```diff
@@ -133,7 +133,11 @@ int CMSketch_IncrBy(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
     RedisModule_ReplyWithArray(ctx, pairCount);
     for (int i = 0; i < pairCount; ++i) {
         size_t count = CMS_IncrBy(cms, pairArray[i].key, pairArray[i].keylen, pairArray[i].value);
-        RedisModule_ReplyWithLongLong(ctx, (long long)count);
+        if (count != UINT32_MAX) {
+            RedisModule_ReplyWithLongLong(ctx, (long long)count);
+        } else {
+            RedisModule_ReplyWithError(ctx, "CMS: INCRBY overflow");
+        }
```
gkorland, line 140:
```suggestion
        if (count == UINT32_MAX) {
            RedisModule_ReplyWithError(ctx, "CMS: INCRBY overflow");
        } else {
            RedisModule_ReplyWithLongLong(ctx, (long long)count);
        }
```

tests/flow/test_cms.py
```diff
@@ -177,6 +177,23 @@ def test_merge_extensive(self):
...
+        
+        self.cmd('FLUSHALL')
+        self.cmd('cms.initbydim', 'cms', '1000', '5')
+        self.assertEqual([large_val], self.cmd('cms.incrby', 'cms', 'a', large_val))
+        self.assertEqual([large_val], self.cmd('cms.query', 'cms', 'a'))
+        self.assertEqual([large_val * 2], self.cmd('cms.incrby', 'cms', 'a', large_val))
+        self.assertEqual([large_val * 2], self.cmd('cms.query', 'cms', 'a'))
+        
+        # overflow as result > UNIT32_MAX
+        res = self.cmd('cms.incrby', 'cms', 'a', large_val)
```
gkorland, line 191: @ashtul please test with few tuples too.

```suggestion
        res = self.cmd('cms.incrby', 'cms', 'a', large_val, 'b', small_val, 'c', large_val)
```

src/cms.c
gkorland: Should you stop the loop after the first error?
If you return an error should you not leave the counter in its original form?

gkorland (changes requested):

src/cms.c
ashtul: @gkorland I don't think we have to stop the loop. This means invalidating the sketch.
In future queries, since we take the minimum value, the sketch can still be used.

As for the order in condition, my preference is to have the common case before the exception. WDYT?

tests/flow/test_cms.py
ashtul: added additional elements

src/cms.c
```diff
@@ -52,11 +47,16 @@ size_t CMS_IncrBy(CMSketch *cms, const char *item, size_t itemlen, size_t value)
     assert(item);
 
     size_t minCount = (size_t)-1;
+    int overflow = 0;
```
gkorland, line 50:
```suggestion
```

gkorland (changes requested):

filipecosta90 (approved):

PR #424, opened 2022-03-08
Migrate doc content to new format

ashtul:
Adjust commands.json and each command .md file to the new format.

MOD-2696


oshadmi (approved):
👍🏼

PR #429, opened 2022-03-20
Enabled CI profilers

filipecosta90:
This PR enables a CI profiler setup with the same specs as the default benchmark setups. 
It attaches a profiler for 60 seconds to each benchmark run and produces a set of outputs by default that can be consulted at 
[CI profiler viewer Grafana dashboard](https://benchmarksrediscom.grafana.net/d/uRPZar57k/ci-profiler-viewer?orgId=1):

- Flame Graph
- Main THREAD Flame Graph
- perf report per dso
- perf report per dso,sym
- perf report per dso,sym with callgraph
- perf report per dso,sym,srcline
- perf report per dso,sym,srcline with callgraph
- perf report top self-cpu
- perf report top self-cpu (dso=)
- Identical stacks collapsed

Note: the results produced with a profiler attached are not used on the visualizations/regression analysis due to the profiler overhead impact.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -95,6 +95,18 @@ commands:
       module_path:
         type: string
         default: ../../redisbloom.so
+      profile_env:
+        type: string
+        default: "0"
+      benchmark_glob:
+        type: string
+        default: "*.yml"
+      triggering_env:
+        type: string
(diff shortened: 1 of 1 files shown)


ashtul (approved):
LGTM

PR #437, opened 2022-04-15
CF.ADDNX item is not multiple

liuchong:
From: https://github.com/RedisBloom/RedisBloom/pull/435

https://redis.io/commands/cf.addnx/

image


chayim (approved):
Thanks @liuchong for fixing this!

ashtul (approved):

PR #452, opened 2022-05-17
Added support for trimmed mean calculation: TDIGEST.TRIMMED_MEAN

filipecosta90:
The following PR adds support for trimmed mean calculation via the new command:
```
TDIGEST.TRIMMED_MEAN {key} {low_cut_percentile} {high_cut_percentile}
```
i.e. return the mean of the data set without the values below and above the low-cut and high-cut percentiles respectively.

Fixes #386 

Note: It implies the review of https://github.com/RedisBloom/t-digest-c/pull/22

PR #455, opened 2022-05-22
Adds man page for tdigest.trimmed_mean

itamarhaber:
Missing from #452 - please verify correctness (especially the return for an empty sketch and example, but also the rest :))


filipecosta90 (approved):
thank you @itamarhaber !

rafie (approved):

Issue #488, opened 2022-07-31
proposal: Add bf.Card command for bloom filter


MakDon:
`BF.CARD` returns the cardinality (number of items inserted) of the bloom filter  stored at key.

## Usage

```
redis> BF.ADD item1 item_foo
(integer) 1
redis9> BF.CARD item1
(integer) 1
redis9> BF.CARD item_new
(integer) 0
```

## Why not BF.INFO

- BF.INFO returns all the info about the filter. But BF.CARD only returns the cardinality.
- BF.INFO returns error when key is not existed. But BF.CARD return 0. (just like [scard][scard], [zcard][zcard] does)

## Why not name it as "bf.count"

In [Another issue](#389 ), @simonprickett suppose to use a command called `bf.count`.    
But there's a command call [cf.count][cf.count], which `Returns the number of times an item may be in the filter`
So it maybe confusing that we use `bf.count` here to return the cardinality of the filter.

## When we need BF.CARD

When we want to use BF as a estimated ZSET.

Assume that some users bought some items, and we want to count "how many users bought a SKU":
```
UserA bought SKU-A, SKU-B, SKU-C,
UserB bought SKU-B
UserC bought SKU-A, SKU-C
UserD bought SKU-D
...
```

And we use the SKU-ID as key, and insert the userID into `how:many:users:bought:SKU-ID`
Then walk through all the SKUs to get the cardinality:
```
BF.CARD how:many:users:bought:SKU-A
BF.CARD how:many:users:bought:SKU-B
BF.CARD how:many:users:bought:SKU-C
BF.CARD how:many:users:bought:SKU-D
BF.CARD how:many:users:bought:SKU-E
```

If we use `bf.INFO` instead of `bf.CARD`, when no one bought SKU-E, it will return error key not exist instead of 0
So we have to check the existence before we use `bf.INFO`.       
`bf.CARD` returns 0 so we no need to `EXIST key` before we query the cardinality.

## Why not BF.INFO {key} [CAPACITY|SIZE|FILTERS|ITEMS|

It is a good idea, but `bf.INFO` returns error when key not exists.
For backward compatibility, I propose use `bf.CARD` instead.

[scard]: https://redis.io/commands/scard/
[zcard]: https://redis.io/commands/zcard/
[cf.count]: https://redis.io/commands/cf.count/

ashtul:
@MakDon Thank you for the proposal.
[...]

PR #489, opened 2022-07-31
Add command bf.card that returns the cardinality (number of items ins…

MakDon:
This PR implements #488

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -197,6 +197,18 @@
     "since": "1.0.0",
     "group": "bf"
   },
+  "BF.CARD": {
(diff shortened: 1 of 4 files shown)


LiorKogan:
@MakDon thank you for your contribution.
We decide not to accept it.
We consider adding optional parameters to `TS.INFO` instead of introducing a new command, 
so you could execute `TS.INFO key CARD` for example, to retrieve only the cardinality.
This addition is planned for all RedisBloom sketch data structures.
As a philosophy, we prefer to avoid introducing more commands than necessary.

PR #503, opened 2022-08-05
Use RM_String in TDIGEST.QUANTILE  reply

ashtul:
remove optimization of `qsort` which allowed for a faster query of quantiles but `double` reply might be different than the user's input. Instead, use the original RM_String as a return value.

### DIFF
--- a/src/rm_tdigest.c
+++ b/src/rm_tdigest.c
@@ -410,13 +410,18 @@ int TDigestSketch_Quantile(RedisModuleCtx *ctx, RedisModuleString **argv, int ar
             return RedisModule_ReplyWithError(ctx, "ERR T-Digest: error parsing quantile");
         }
     }
-    qsort(quantiles, n_quantiles, sizeof(double), double_cmpfunc);
     double *values = (double *)__td_calloc(n_quantiles, sizeof(double));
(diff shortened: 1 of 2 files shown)


filipecosta90 (dismissed):

src/rm_tdigest.c
```diff
@@ -410,13 +410,18 @@ int TDigestSketch_Quantile(RedisModuleCtx *ctx, RedisModuleString **argv, int ar
             return RedisModule_ReplyWithError(ctx, "ERR T-Digest: error parsing quantile");
         }
     }
-    qsort(quantiles, n_quantiles, sizeof(double), double_cmpfunc);
     double *values = (double *)__td_calloc(n_quantiles, sizeof(double));
-    td_quantiles(tdigest, quantiles, values, n_quantiles);
+    for (int i = 0; i < n_quantiles; ++i) {
+        int start = i;
+        while (i < n_quantiles - 1 && quantiles[i] < quantiles[i + 1]) {
```
LiorKogan, line 416: Maybe quantiles[i] `<=` quantiles[i + 1]) ?

LiorKogan (commented):
suggested change

PR #509, opened 2022-08-09
Update commands.json

shacharPash:
In TOPK.LIST - Change WITHCOUNT paraneter to optional.

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -703,6 +703,7 @@
         "name": "withcount",
         "token": "WITHCOUNT",
         "type": "pure-token"
+        "optional": true
       }
     ],
     "since": "2.0.0",


gkorland (approved):

commands.json
```diff
@@ -703,6 +703,7 @@
         "name": "withcount",
         "token": "WITHCOUNT",
         "type": "pure-token"
```
romansergeenkosofteq, line 705: hi, you missed comma here

itamarhaber: Yep - breaks my redis.io build :)

itamarhaber: Here's the fix: https://github.com/RedisBloom/RedisBloom/pull/510

PR #512, opened 2022-08-10
mark TOPK.COUNT as deprecated

ashtul:


LiorKogan (commented):
People will ask why is it deprecated.
I think we should update TOPK.COUNT documentation as well, and describe what are the limitations of this function, and also mention specifically that this function will be deprecated, and mention any alternatives (if there are any).

PR #530, opened 2022-09-08
Introduce TDIGEST.RANK and TDIGEST.REVRANK commands

OfirMos:
# 1. TDIGEST.RANK

 Introduce `TDIGEST.RANK key value [value...]` Retrieve the estimated rank of value (the number of observations in the sketch that are smaller than value + half the number of observations that are equal to value).

Return (processed according to this order): 

 -   Nan if the sketch is empty

  -  Error if value is NaN

   -  Integer in [-1 .. n] where n is the sum of weights added to the sketch
    (-1 when value < value of the smallest observation, n when value > value of the largest observation).

# 2. TDIGEST.REVRANK

 Introduce `TDIGEST.REVRANK key value [value...]` Retrieve the estimated rank of value (the number of observations in the sketch that are larger than value + half the number of observations that are equal to value).

Return (processed according to this order): 

 -   Nan if the sketch is empty

  -  Error if value is NaN

   -  Integer in [-1 .. n] where n is the sum of weights added to the sketch
    (-1 when value > value of the smallest observation, n when value < value of the smallest observation).


src/rm_tdigest.c
```diff
@@ -389,6 +389,73 @@ static int double_cmpfunc(const void *a, const void *b) {
...
+    td_histogram_t *tdigest = RedisModule_ModuleTypeGetValue(key);
+    if(td_size(tdigest) == 0) {
+        RedisModule_ReplyWithArray(ctx, 1);
+        RedisModule_ReplyWithDouble(ctx, NAN);
+        RedisModule_CloseKey(key);
+        return;
+    }
+
+    const size_t n_values = argc - 2;
+    double *vals = (double *)__td_calloc(n_values, sizeof(double));
```
ashtul, line 424: malloc

src/rm_tdigest.c
```diff
@@ -389,6 +389,73 @@ static int double_cmpfunc(const void *a, const void *b) {
...
+            __td_free(vals);
+            return RedisModule_ReplyWithError(ctx, "ERR T-Digest: error parsing value");
+        }
+    }
+
+    int *ranks = (int *)__td_calloc(n_values, sizeof(int));
+
+    double size = td_size(tdigest);
+    double min = td_min(tdigest);
+    double max = td_min(tdigest);
```
ashtul, line 438:
```suggestion
    double max = td_max(tdigest);
```

src/rm_tdigest.c
```diff
@@ -389,6 +389,73 @@ static int double_cmpfunc(const void *a, const void *b) {
...
+    for (int i = 0; i < n_values; ++i) {
+        if(vals[i] < min) {
+            ranks[i] = -1;
+        } else if (vals[i] > max) {
+            ranks[i] = td_size(tdigest);
```
ashtul, line 443: What does `-1` or `td_size` symbol?

ashtul (commented):
A few questions
Also please add to `commands.json`

src/rm_tdigest.c
filipecosta90: fixed in the latest comming

src/rm_tdigest.c
filipecosta90: I've updated the PR description to include the command behavior:
 (-1 when value < value of the smallest observation, n when value > value of the largest observation).

src/rm_tdigest.c
filipecosta90: @ashtul we can reach the same outcome with calloc and the overhead for such small array is neglectable IMHO

ashtul (approved):
LGTM

PR #531, opened 2022-09-08
TDIGEST.ADD value and weight checks (As in the reference implementation)

filipecosta90:
- Allow only positive integer weights on TDIGEST.ADD (As in the reference implementation)
- Disallow INF and -INF in TDIGEST.ADD


ashtul (approved):

LiorKogan (approved):

ashtul (approved):
LGTM

PR #532, opened 2022-09-08
TDIGEST.MERGE and TDIGEST.MERGESTORE are now a single command. TDIGEST.MERGE can now merge to itself.

filipecosta90:


src/rm_tdigest.c
```diff
@@ -604,7 +606,7 @@ void *TDigestRdbLoad(RedisModuleIO *rdb, int encver) {
     const double compression = RedisModule_LoadDouble(rdb);
     td_histogram_t *tdigest = td_new(compression);
     tdigest->min = RedisModule_LoadDouble(rdb);
-    tdigest->min = RedisModule_LoadDouble(rdb);
+    tdigest->max = RedisModule_LoadDouble(rdb);
```
ashtul, line 609: Do we have any tests to cover RDB?
If not, lets add them in another PR.

tests/flow/test_tdigest.py
```diff
@@ -170,79 +170,123 @@ def test_negative_tdigest_add(self):
...
-    def test_tdigest_mergestore_percentile(self):
+        # merge to a t-digest that already exists with non-default compression
+        self.assertOk(self.cmd("tdigest.merge", "to-tdigest-50", "2","from-1", "from-2", "COMPRESSION", "500"))
+        # ensure tha the destination t-digest has the passed compression
+        to_info = parse_tdigest_info(self.cmd("tdigest.info", "to-tdigest-50"))
+        compression = int(to_info["Compression"])
+        self.assertEqual(500, compression)
+
+        # merge to a t-digest that already exists but given we specify override it will use the max compression of all inputs
+        self.assertOk(self.cmd("tdigest.merge", "to-tdigest-50", "2","from-1", "from-2", "OVERRIDE"))
```
ashtul, line 276: Can we override the same key or then the result will be the same?
`self.assertOk(self.cmd("tdigest.merge", "from-1, "2","from-1", "from-2", "OVERRIDE"))`

src/rm_tdigest.c
```diff
@@ -264,33 +231,61 @@ int TDigestSketch_MergeStore(RedisModuleCtx *ctx, RedisModuleString **argv, int
         goto cleanup;
     }
     long long compression = 0;
-    // If no compression value is passed, the used compression will the maximal value amongst all
-    // inputs.
-    bool use_max_compression = true;
+    if (to_exists) {
+        compression = tdigestToStart->compression;
+    }
```
ashtul, line 236:
```suggestion
    long long compression = to_exists ? tdigestToStart->compression : 0;
```

ashtul (dismissed):
Looks good.
minor suggestion

tests/flow/test_tdigest.py
filipecosta90: we can override. I've added a specific test to illustrate that: 
https://github.com/RedisBloom/RedisBloom/blob/a877bfe9ca3b717a6abba945446945f2bfb4ee23/tests/flow/test_tdigest.py#L193

src/rm_tdigest.c
filipecosta90: I believe we don't test enough the rdb load otherwise we should have caught this before. 
I've reverted TDigestRdbLoad fix given its not related to this PR (addressing separately)

ashtul (approved):

PR #535, opened 2022-09-09
TDIGEST.TRIMMED_MEAN: enforce low_cut_percentile to be lower than high_cut_percentile (as in reference implementations)

filipecosta90:
Given the original tdigest does not have a reference implementation:

- Compared implementation vs postgresql tdigest trimmed mean implementation: https://github.com/tvondra/tdigest/blob/master/tdigest.c
- Confirmed also that using google sheets the result of TRIMMEAN function results in the same output of our implementation.

### DIFF
--- a/tests/flow/test_tdigest.py
+++ b/tests/flow/test_tdigest.py
@@ -507,7 +507,7 @@ class testTDigest:
 
     def test_tdigest_trimmed_mean(self):
         self.cmd('FLUSHALL')
-        self.assertOk(self.cmd("tdigest.create", "tdigest", "compression", 500))
+        self.assertOk(self.cmd("tdigest.create", "tdigest", "compression", 1000))
         # insert datapoints into sketch
         for x in range(0, 20):
             self.assertOk(self.cmd("tdigest.add", "tdigest", x, 1.0))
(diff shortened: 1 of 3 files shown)


tests/flow/test_tdigest.py
```diff
@@ -521,6 +521,45 @@ def test_tdigest_trimmed_mean(self):
...
+            )
+        self.assertAlmostEqual(
+                1.4, float(self.cmd("tdigest.trimmed_mean", "tdigest", 0.095, 0.905)), 0.01
+            )
+        self.assertAlmostEqual(
+                1.4, float(self.cmd("tdigest.trimmed_mean", "tdigest", 0.05, 0.95)), 0.01
+            )
+        self.assertAlmostEqual(
+                1.4, float(self.cmd("tdigest.trimmed_mean", "tdigest", 0.01, 0.99)), 0.01
+            )
```
ashtul, line 562: Does trim-mean has to be symmetric on both sides?

filipecosta90: Nope. I've added those tests given it was simpler to produce the expected outcomes.

ashtul (approved):
LGTM

PR #545, opened 2022-09-19
 Ensure allocation check on TDIGEST.CREATE and TDIGEST.MERGE

filipecosta90:
Before we did not check that the allocation was indeed successful of the underlying data structure. This PR fixes it and adds tests to confirm we're indeed replying with an error in the case of failed allocations.

### DIFF
--- a/src/rm_tdigest.c
+++ b/src/rm_tdigest.c
@@ -100,7 +100,10 @@ int TDigestSketch_Create(RedisModuleCtx *ctx, RedisModuleString **argv, int argc
             return REDISMODULE_ERR;
         }
     }
-    tdigest = td_new(compression);
+    if (td_init(compression, &tdigest) != 0) {
+        RedisModule_CloseKey(key);
+        return RedisModule_ReplyWithError(ctx, "ERR T-Digest: allocation failed");
+    }
     if (RedisModule_ModuleTypeSetValue(key, TDigestSketchType, tdigest) != REDISMODULE_OK) {
         RedisModule_CloseKey(key);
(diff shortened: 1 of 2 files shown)


src/rm_tdigest.c
```diff
@@ -100,7 +100,10 @@ int TDigestSketch_Create(RedisModuleCtx *ctx, RedisModuleString **argv, int argc
             return REDISMODULE_ERR;
         }
     }
-    tdigest = td_new(compression);
```
ashtul, line 103: Why not leave the current behavior but check the pointer is not NULL?

src/rm_tdigest.c
```diff
@@ -306,7 +309,10 @@ int TDigestSketch_Merge(RedisModuleCtx *ctx, RedisModuleString **argv, int argc)
             compression = td_compression > compression ? td_compression : compression;
         }
     }
-    tdigestTo = td_new(compression);
```
ashtul, line 309: Same

src/rm_tdigest.c
filipecosta90: @ashtul the td_init() API will allow us to be more verbose on the error in the future if we want based on the reply value. If you agree lets use td_init instead of td_new. Furthermore td_new just simply calls td_init so it's one less function call.

src/rm_tdigest.c
filipecosta90: see comment aboe

ashtul (approved):
LGTM

PR #578, opened 2022-10-25
Fixes per TDIGEST.RANK empty reply and rounding errors. TDIGEST.ADD input check fixes

filipecosta90:
- Fixed TDIGEST.RANK empty reply
- Fixed TDIGEST.ADD input ingestion with mixed bad inputs
- Addressed rounding error of rank


OfirMos (approved):

ashtul (approved):
LGTM

Issue #389, opened 2022-01-13
Add a way of getting to the count / size / cardinality of the Bloom Filter directly


simonprickett:
I can get information about how many items were inserted into the Bloom Filter from `BF.INFO` by parsing the result out:

```bash
127.0.0.1:6379> bf.madd somebloom the quick brown fox ran up the hill and did some stuff because it can
 1) (integer) 1
 2) (integer) 1
 3) (integer) 1
 4) (integer) 1
 5) (integer) 1
 6) (integer) 1
 7) (integer) 0
 8) (integer) 1
 9) (integer) 1
10) (integer) 1
11) (integer) 1
12) (integer) 1
13) (integer) 1
14) (integer) 1
15) (integer) 1
127.0.0.1:6379> bf.info somebloom
 1) Capacity
 2) (integer) 100
 3) Size
 4) (integer) 290
 5) Number of filters
 6) (integer) 1
 7) Number of items inserted
 8) (integer) 14
 9) Expansion rate
10) (integer) 2
```

I feel it would be useful to have a command that did something like this (equivalent to `PFCOUNT` for Hyperloglog):

```bash
127.0.0.1:6379> bf.count somebloom
(integer) 14
```

LiorKogan:
Thank you @simonprickett 

Maybe we should add optional flags - for receiving a single value.
`BF.INFO {key} [CAPACITY|SIZE|FILTERS|ITEMS|ERATE]`

What do you think?

simonprickett:
@LiorKogan yes that would make sense for me too.  Thanks for looking at this.

MakDon:
Want this feature too.
Maybe it would be nice that `bf.count not:exist:key` returns `0` instead of an error `-ERR not found`

MakDon:
Hi @LiorKogan , could we add an option here, with which `BF.INFO [CAPACITY|SIZE|FILTERS|ITEMS|EXPANSION]` return 0 instead of `-ERR not found`?  Otherwise, we have to check if the key exists before `BF.INFO`, which would waste a RTT and is not convenient for pipelining.
The scene has been in #488 in detail.
```
127.0.0.1:6379> BF.INFO key:not:exist
-ERR not found
127.0.0.1:6379> BF.INFO key:not:exist NC
 1) (integer) 0
```
Thanks very much.

LiorKogan:
@MakDon this goes against Redis philosophy.
Commands that receive a key should error when the key doesn't exist or of the wrong data type.

[...]

PR #604, opened 2023-01-02
Add new command BF.CARD

MakDon:
As discussion in #389 ,  we introduce a new command `BF.CARD`, which returns the cardinality (number of items inserted) of the bloom filter stored at key.
As mentioned in #389 , `BF.CARD`:
> Return 0 when key does not exist.
> Error when key is of a type other than Bloom filter.
> Note: when key exists - return the same value as BF.INFO key ITEMS.

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -229,6 +229,18 @@
     "since": "1.0.0",
     "group": "bf"
   },
+  "BF.CARD": {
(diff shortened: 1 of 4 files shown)


ashtul (approved):
Looks great. 
Thank you.

PR #610, opened 2023-01-18
[BUG] SubCF->numBuckets change to uint64

### LINKED ISSUE #609: Cuckoo Filter memory usage reporting
There doesn't appear to be a way to return how much memory a cuckoo filter is consuming:

```
127.0.0.1:6378[2]> CF.INFO test_filter_cuckoo
 1) Size
 2) (integer) 8589934664
 3) Number of buckets
 4) (integer) 2147483648
 5) Number of filters
 6) (integer) 2
 7) Number of items inserted
 8) (integer) 141679676
 9) Number of items deleted
10) (integer) 0
11) Bucket size
12) (integer) 2
13) Expansion rate
14) (integer) 1
15) Max iterations
16) (integer) 20

127.0.0.1:6378[2]> MEMORY USAGE  test_filter_cuckoo
(integer) 120
127.0.0.1:6378[2]>
```

Bloom filters return their `Size` from `BF.INFO`.  So I was expecting `CF.INFO` to include similar information.  `MEMORY USAGE` does not seem like it could possibly be correct in this case.  My Cuckoo Filter is sized with a capacity of `2^32`.

ashtul:
This is a bug fix for issue #609.
unit_32 was used and get replaced with unit_64 as a 56 bit field which is more than enough for any imaginable system.
size 
Note: The decision to not introduce a new independent variable was in order to avoid a breaking change as the size of the filter in CF.INFO would change.

Fixes #609

### DIFF
--- a/src/cuckoo.h
+++ b/src/cuckoo.h
@@ -25,8 +25,8 @@ typedef uint8_t CuckooBucket[1];
 typedef uint8_t MyCuckooBucket;
 
 typedef struct {
-    uint32_t numBuckets;
-    uint8_t bucketSize;
+    uint64_t numBuckets : 56;
+    uint64_t bucketSize : 8;
     MyCuckooBucket *data;
 } SubCF;
 


gkorland:
@ashtul can you please add a relevant test?

rafie (approved):

PR #613, opened 2023-01-29
[2.4] Cherry-picks for v2.4.4

rafie:


MakDon:
We introduced a new command `BF.CARD` in #604 , which is cherry-picked into this PR.
If we cherry-pick this commit into v2.4.4, it would break [Semantic Versioning](https://semver.org/) rule "PATCH version when you make backwards compatible bug fixes". But IMHO, it doesn't matter.


And we need to update [the release version of BF.CARD](https://github.com/RedisBloom/RedisBloom/blob/145c8fcbae3842689a32405998a6553094c9e946/commands.json#L241) to delacre that this command is available since v2.4.4.

The [note](https://github.com/RedisBloom/RedisBloom/blob/master/docs/commands/bf.card.md#note-this-command-will-be-available-on-the-next-release) in docs should be removed as well.

LiorKogan:
Thank you @MakDon

PR #670, opened 2023-08-09
update web pages for tabbed examples

sav-norem:
to get tabbed examples in CLI and client languages, we need to update each data type page to include hugo shortcodes and updated CLI examples using bicycles for a cohesive examples experience

PR #680, opened 2023-10-05
Update count-min-sketch.md

sav-norem:
update count min sketch with hugo shortcodes and slightly more robust example of the CMS data type


docs/docs/count-min-sketch.md
```diff
@@ -24,11 +24,29 @@ This application answers this question: What was the sales volume (on a certain
...
 
 ## Examples
-Let's say we choose an error of 0.1%(`0.001`) with certainty of 99.8%(`0.998`) (thus probability of error 0.02% (`0.002`)). The resulting sketch will try to keep the error within 0.1% of the sum of counts of **ALL** elements that have been added to the sketch and the probability for this error to be higher  than that (a collision of an element below the threshold with an element above the threshold) will be 0.02%. 
-
-```
-> CMS.INITBYPROB key 0.001 0.002
-```
+Let's say we choose an error of 0.1%(`0.001`) with certainty of 99.8%(`0.998`) (thus probability of error 0.02% (`0.002`)). The resulting sketch will try to keep the error within 0.1% of the sum of counts of **ALL** elements that have been added to the sketch and the probability for this error to be higher than that (a collision of an element below the threshold with an element above the threshold) will be 0.02%. We'll add a few things to the CMS and then check their frequency. Note that like other probablilistic data structures, this small of an example likely won't show any collisions.
```
dwdougherty, line 27:
```suggestion
Assume you select an error rate of 0.1% (0.001) with a certainty of 99.8% (0.998). This means you have an error probability of 0.02% (0.002). Your sketch strives to keep the error within 0.1% of the total count of all elements you've added. There's a 0.02% chance the error might exceed this—like when an element below the threshold overlaps with one above it. When you add a few items to the CMS and evaluate their frequency, remember that in such a small sample, collisions are rare, as seen with other probabilistic data structures.
```

dwdougherty (changes requested):
Some re-wording of the text; took me a while to unwind what this paragraph was trying to convey (not your fault at all, you inherited it). Let me know if the updated wording works for you.

sav-norem:
@dwdougherty reads easier to me! just accepted that suggestion

dwdougherty (approved):
Builds correctly; approved.

sav-norem:
updated master to include all tabbed example changes - closing as superfluous

sav-norem:
merged into PR #670

PR #687, opened 2023-10-11
Update t-digest.md

sav-norem:
update tdigest data type page to include hugo short codes and relevant examples

### DIFF
--- a/docs/docs/t-digest.md
+++ b/docs/docs/t-digest.md
@@ -62,23 +62,17 @@ You measure the IP packets transferred over your network each second and try to 
 
 ## Examples
 
-#### Creating a t-digest
+We'll demonstrate creating a t-digest with an initial compression of 100, adding items to it, and checking the estimated percentile of a value. Note that the `COMPRESSION` argument is used to specify the tradeoff between accuracy and memory consumption. The default is 100. Higher values mean more accuracy. Note also that unlike some of the other probabilistic data structures, the `TDIGEST.ADD` command will not create a new structure if the key does not exist.
 
-```
-> TDIGEST.CREATE my-tdigest COMPRESSION 100
-```
+{{< clients-example tdigest_tutorial tdig_start >}}
+> TDIGEST.CREATE bikes:sales COMPRESSION 100
+OK
+> TDIGEST.ADD bikes:sales 21
+OK
+> TDIGEST.ADD bikes:sales 150 95 75 34
+OK
+{{< /clients-example >}}
 
-The `COMPRESSION` argument is used to specify the tradeoff between accuracy and memory consumption. The default is 100. Higher values mean more accuracy.
-
-#### Adding a single element to the t-digest:
-```
-> TDIGEST.ADD my-tdigest 20.9
-```
-
-#### Adding multiple elements to the t-digest:
-```
-> TDIGEST.ADD my-tdigest 308 315.9
-```
 
 You can repeat calling [TDIGEST.ADD](https://redis.io/commands/tdigest.add/) whenever new observations are available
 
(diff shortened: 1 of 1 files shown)


docs/docs/t-digest.md
```diff
@@ -62,62 +62,56 @@ You measure the IP packets transferred over your network each second and try to
 
 ## Examples
 
-#### Creating a t-digest
+We'll demonstrate creating a t-digest with an initial compression of 100, adding items to it, and checking the estimated percentile of a value. Note that the `COMPRESSION` argument is used to specify the tradeoff between accuracy and memory consumption. The default is 100. Higher values mean more accuracy. Note also that unlike some of the other probabilistic data structures, the `TDIGEST.ADD` command will not create a new structure if the key does not exist.
```
dwdougherty, line 65: I don't see where you "check the estimated percentile of a value" in the example.

```suggestion
In the following example, you'll create a t-digest with an initial compression of 100, add items to it, and check the estimated percentile of a value. The `COMPRESSION` argument is used to specify the tradeoff between accuracy and memory consumption. The default value is 100. Higher values mean more accuracy. Note: unlike some of the other probabilistic data structures, the `TDIGEST.ADD` command will not create a new structure if the key does not exist.
```

docs/docs/t-digest.md
```diff
@@ -62,62 +62,56 @@ You measure the IP packets transferred over your network each second and try to
...
 
 Another helpful feature in t-digest is CDF (definition of rank) which gives us the fraction of observations smaller or equal to a certain value. This command is very useful to answer questions like "*What's the percentage of observations with a value lower or equal to X*".
 
->More precisely, `TDIGEST.CDF` will return the estimated fraction of observations in the sketch that are smaller than X plus half the number of observations that are equal to X
-
-Let's illustrate this with an example: if we have a set of observations of people's age with gaussian distribution, we can ask a question like "What's the percentage of people younger than 50 years?"
+>More precisely, `TDIGEST.CDF` will return the estimated fraction of observations in the sketch that are smaller than X plus half the number of observations that are equal to X. We can also use the `TDIGEST.RANK` command, which is very similar. Instead of returning a fraction, it returns the **number** of observations in the sketch that are smaller than X plus half the number of observations that are equal to X, or in other words - the estimated rank of a value. The `TDIGEST.RANK` command is also variadic.
```
dwdougherty, line 83:
```suggestion
>More precisely, `TDIGEST.CDF` will return the estimated fraction of observations in the sketch that are smaller than X, plus half the number of observations that are equal to X. We can also use the `TDIGEST.RANK` command, which is very similar. Instead of returning a fraction, it returns the *number* of observations in the sketch that are smaller than X, plus half the number of observations that are equal to X, or, in other words, the estimated rank of a value. The `TDIGEST.RANK` command is also variadic.
```

dwdougherty (changes requested):
A few suggested changes.

docs/docs/t-digest.md
LiorKogan: I wouldn't use "initial compression" because it suggests that it can be changed or is being changed automatically. I would use just "compression" instead. 
I also suggest changing "check the estimated percentile" to "query ...".

docs/docs/t-digest.md
LiorKogan: Please add "estimated":
> Instead of returning a fraction, it returns the ----estimated---- **number** of observations.

> The `TDIGEST.RANK` command is also variadic.
Consider using a simpler explanation. I'm not sure every reader knows what "variadic" means here.

docs/docs/t-digest.md
```diff
@@ -62,62 +62,56 @@ You measure the IP packets transferred over your network each second and try to
...
 
-```
-> TDIGEST.ADD my-tdigest 45.88 44.2 58.03 19.76 39.84 69.28 50.97 25.41 19.27 85.71 42.63
+Let's illustrate this with an example: if we have a set of observations of people's age with gaussian distribution, we can ask a question like "What's the percentage of bike racers are younger than 50 years?"
```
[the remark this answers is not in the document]
LiorKogan: We don't need the Gaussian distribution assumption. t-digest works well with any distribution.

docs/docs/t-digest.md
```diff
@@ -141,23 +135,28 @@ If `destKey` is an existing sketch, its values are merged with the values of the
...
-> TDIGEST.MAX my-tdigest
-```
+{{< clients-example tdigest_tutorial tdig_min >}}
+> TDIGEST.MIN racer_ages
+"19.27"
+> TDIGEST.MAX racer_ages
+"85.709999999999994"
+{{< /clients-example >}}
 
 Both return nan when the sketch is empty.
```
LiorKogan, line 145: change nan to `nan` ?

sav-norem:
@dwdougherty @LiorKogan I've just pushed a change to reflect all these comments. I simplified some language and clarified other. Great feedback, let me know if there's more!

sav-norem:
Closing as wrapped this file into Bloom filter PR - https://github.com/RedisBloom/RedisBloom/pull/670

PR #688, opened 2023-10-12
Update top-k.md

sav-norem:
update top-k data type page to include hugo short codes - condense example and add context


sav-norem:
@dwdougherty thinking you may have missed this one

docs/docs/top-k.md
```diff
@@ -41,38 +41,43 @@ The `TOPK.LIST` command has a time complexity of `O(K)` so if `K` is small, ther
...
-```
-> TOPK.ADD my-topk foo bar 42
-```
-
-* Return list of the top K items
-
-```
-> TOPK.LIST my-topk 
-```
+* We'll demonstrate initializing a Top-K with specific parameters; though `width`, `depth`, and `decay_constant` can be left out and default values will be used.
```
dwdougherty, line 44:
```suggestion
In the following example, you will initialize a Top-K with specific parameters. If `width`, `depth`, and `decay_constant` are not provided, default values will be used.
```

dwdougherty:
I edited this page and I sent you the whole thing in a Slack message on Mon. Oct. 16 @ ~1:35 pm PDT.

docs/docs/top-k.md
```diff
@@ -14,7 +14,7 @@ One very common application is detecting network anomalies and DDoS attacks wher
...
+The Redis Stack implementation of Top-K is based on the [HeavyKeepers](https://www.usenix.org/conference/atc18/presentation/gong) algorithm presented by Junzhi Gong et al. It discards some older approaches like "count-all" and "admit-all-count-some" in favour of a "**count-with-exponential-decay**" strategy which is biased against mouse (small) flows and has a limited impact on elephant (large) flows. This implementation involves using two data structures in tandem; a hash table to hold the probabilitic counts (much like the Count-Min Sketch), along with a min heap that holds the `K` items with the highest counts. This ensures high accuracy with shorter execution times than previous probabilistic algorithms allowed, while keeping memory utilization to a fraction of what is typically required by a Sorted Set and also having the additional benefit of being able to get real time notifications when elements are added or removed from the Top K list.
```
dwdougherty, line 17:
```suggestion
The Redis Stack implementation of Top-K is based on the [HeavyKeepers](https://www.usenix.org/conference/atc18/presentation/gong) algorithm presented by Junzhi Gong et al. It discards some older approaches like "count-all" and "admit-all-count-some" in favour of a "**count-with-exponential-decay**" strategy which is biased against mouse (small) flows and has a limited impact on elephant (large) flows. This implementation uses two data structures in tandem: a hash table that holds the probabilistic counts (much like the Count-Min Sketch), and a min heap that holds the `K` items with the highest counts. This ensures high accuracy with shorter execution times than previous probabilistic algorithms allowed, while keeping memory utilization to a fraction of what is typically required by a Sorted Set. It has the additional benefit of being able to get real time notifications when elements are added or removed from the Top K list. 
```

dwdougherty (changes requested):
Just a coupla more edits.

dwdougherty (approved):
LGTM; builds correctly.

docs/docs/top-k.md
```diff
@@ -26,7 +26,7 @@ Data flow is the incoming game scores. Flow id is the user id, and the value is
 Every time a user scores points, they're added to the Top K list. 
 
 - If the result is `nil`, the user is already in the Top K. Update the sorted set with the user's new score. 
-- if the result is an id of another player, the player you just added (`id1`) took over the player who got returned (`id2`). Remove player `id2` from the sorted set and add player `id1`. 
+- If the result is an id of another player, the player you just added (`id1`) took over the player who got returned (`id2`). Remove player `id2` from the sorted set and add player `id1`.
```
LiorKogan, line 29: Why do you need to maintain a separate sorted set? Why not call `TOPK.LIST`?

sav-norem: I didn't actually write this part and was just editing the if -> If, but can definitely reword this bit as well.

sav-norem: Seems like the idea was to maintain both a top-k and a full on leaderbaord with a sorted set so that no matter how many players are added, you can still see their scores. If you're not in the top-k players, I don't think there's a call to retrieve your score. Well, there's `count` but it's deprecated so I definitely don't want to recommend that in the docs.

LiorKogan: Well, I guess it is not a good use case for top-k if you want to retrieve the score. As for the current example - it seems redundant to use top-k as well.

PR #690, opened 2023-10-17
fix tests on redis cluster: add shard id to the keys

OfirMos:


oshadmi (approved):
Approved assuming  no cross-key is handled by Bloom and only keys on the same shard are involved

PR #718, opened 2023-12-28
Add setuptools package to python dependencies to fix MacOS build

tezc:
```
    File "/Users/runner/Library/Python/3.12/lib/python/site-packages/RLTest/__init__.py", line 3, in <module>
      from ._version import __version__
    File "/Users/runner/Library/Python/3.12/lib/python/site-packages/RLTest/_version.py", line 4, in <module>
      import pkg_resources
  ModuleNotFoundError: No module named 'pkg_resources'
```
MacOS build on github actions fails with above error: [link](https://github.com/RedisBloom/RedisBloom/actions/runs/7284882607/job/19850896589)

Looks like RLTest needs `setuptools` package and it does not come by default on this macOS build.

### DIFF
--- a/tests/flow/requirements.txt
+++ b/tests/flow/requirements.txt
@@ -1,2 +1,5 @@
+# MacOS setup on github actions needs setuptools (required by RLTest).
+# Normally, it would be better RLTest have this package as dependency.
+setuptools
 RLTest == 0.7.5
 numpy


tezc:
@alonre24 maybe you can suggest a better way to handle this one. Should/can we add `setuptools` as dependency to RLTest?

tests/flow/requirements.txt
```diff
@@ -1,2 +1,5 @@
+# MacOS setup on github actions needs setuptools (required by RLTest).
+# Normally, it would be better RLTest have this package as dependency.
+setuptools
```
YaacovHazan, line 3: Please move it to `./install/mariner2.sh` (`pip install --upgrade setuptools`) you can see what I did in `1.10`.
btw I think there are more issue with macos that we need to cp from `1.10`:
`2c418b0 2023-12-25 change mac-os-64 to 13.4.1 (YaacovHazan)`
`d2dfbc3 2023-12-30 fix macos python installation (YaacovHazan)`

but let's wait with it as I have a conversation about it with CTO...

YaacovHazan (changes requested):

PR #724, opened 2024-01-09
MOD-6343 Fix potential crash for cf.reserve

tezc:
Currently, there is no upper bound check for cf.reserve arguments BUCKETSIZE, MAXITERATIONS and EXPANSION. 
Accepting a large value for these arguments can cause a crash due to overflow e.g.:

```
CF.RESERVE CF 67108864 BUCKETSIZE 33554432
```

Deciding upper limits for these arguments:

```
typedef struct {
    uint64_t numBuckets : 56;
    uint64_t bucketSize : 8;
    MyCuckooBucket *data;
} SubCF;

typedef struct {
    ....
    uint16_t bucketSize;
    uint16_t maxIterations;
    uint16_t expansion;
    SubCF *filters;
} CuckooFilter;
```

- For BUCKETSIZE, we store it in 8 bits inside struct SubCF. It makes upper limit for BUCKETSIZE: 255
- MAXITERATIONS is stored as uint16_t which makes upper limit: 65535
- EXPANSION is stored as uint16_t as well but it is required to be power of two: 32768

Although it seems algorithm can handle higher upper limits, this PR does not increase
storage for these arguments not to increase memory usage for the most common cases.

**Potentially breaking change**
Currently, if users are passing higher values than these limits, it might still work. 
Though, it is not a well-defined behavior and these usages might have bugs in it. 
After this PR, these cf.reserve calls will get error reply.

### DIFF
--- a/src/rebloom.c
+++ b/src/rebloom.c
@@ -541,8 +541,8 @@ static int CFReserve_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv,
         if (RedisModule_StringToLongLong(argv[mi_loc + 1], &maxIterations) != REDISMODULE_OK) {
             return RedisModule_ReplyWithError(ctx, "Couldn't parse MAXITERATIONS");
         } else if (maxIterations <= 0 || maxIterations > CF_MAX_ITERATIONS) {
-            return RedisModule_ReplyWithError(
-                ctx, "MAXITERATIONS parameter must be between 1 and 65535 (both are inclusive).");
+            return RedisModule_ReplyWithError(ctx, "MAXITERATIONS parameter must be an integer "
+                                                   "between 1 and 65535 (both are inclusive).");
         }
     }
 
@@ -553,7 +553,8 @@ static int CFReserve_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv,
             return RedisModule_ReplyWithError(ctx, "Couldn't parse BUCKETSIZE");
         } else if (bucketSize <= 0 || bucketSize > CF_MAX_BUCKET_SIZE) {
             return RedisModule_ReplyWithError(
-                ctx, "BUCKETSIZE parameter must be between 1 and 255 (both are inclusive).");
+                ctx,
+                "BUCKETSIZE parameter must be an integer between 1 and 255 (both are inclusive).");
         }
     }
 
@@ -564,7 +565,8 @@ static int CFReserve_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv,
             return RedisModule_ReplyWithError(ctx, "Couldn't parse EXPANSION");
         } else if (expansion < 0 || expansion > CF_MAX_EXPANSION) {
             return RedisModule_ReplyWithError(
-                ctx, "EXPANSION parameter must be between 0 and 32768 (both are inclusive).");
+                ctx,
+                "EXPANSION parameter must be an integer between 0 and 32768 (both are inclusive).");
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/rebloom.c
```diff
@@ -559,9 +562,9 @@ static int CFReserve_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv,
     if (ex_loc != -1) {
         if (RedisModule_StringToLongLong(argv[ex_loc + 1], &expansion) != REDISMODULE_OK) {
             return RedisModule_ReplyWithError(ctx, "Couldn't parse EXPANSION");
-        } else if (expansion < 0) {
+        } else if (expansion < 0 || expansion > CF_MAX_EXPANSION) {
             return RedisModule_ReplyWithError(
-                ctx, "EXPANSION parameter needs to be a non-negative integer");
+                ctx, "EXPANSION parameter must be between 0 and 32768 (both are inclusive).");
```
LiorKogan, line 567: BUCKETSIZE must be an integer between 0 and 32768

src/rebloom.c
```diff
@@ -532,14 +535,14 @@ static int CFReserve_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv,
...
+    long long maxIterations = CF_DEFAULT_MAX_ITERATIONS;
     int mi_loc = RMUtil_ArgIndex("MAXITERATIONS", argv, argc);
     if (mi_loc != -1) {
         if (RedisModule_StringToLongLong(argv[mi_loc + 1], &maxIterations) != REDISMODULE_OK) {
             return RedisModule_ReplyWithError(ctx, "Couldn't parse MAXITERATIONS");
-        } else if (maxIterations <= 0) {
+        } else if (maxIterations <= 0 || maxIterations > CF_MAX_ITERATIONS) {
             return RedisModule_ReplyWithError(
-                ctx, "MAXITERATIONS parameter needs to be a positive integer");
+                ctx, "MAXITERATIONS parameter must be between 1 and 65535 (both are inclusive).");
```
LiorKogan, line 545: MAXITERATIONS must be an integer between 1 and 65535

src/rebloom.c
```diff
@@ -548,9 +551,9 @@ static int CFReserve_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv,
     if (bs_loc != -1) {
         if (RedisModule_StringToLongLong(argv[bs_loc + 1], &bucketSize) != REDISMODULE_OK) {
             return RedisModule_ReplyWithError(ctx, "Couldn't parse BUCKETSIZE");
-        } else if (bucketSize <= 0) {
+        } else if (bucketSize <= 0 || bucketSize > CF_MAX_BUCKET_SIZE) {
             return RedisModule_ReplyWithError(
-                ctx, "BUCKETSIZE parameter needs to be a positive integer");
+                ctx, "BUCKETSIZE parameter must be between 1 and 255 (both are inclusive).");
```
LiorKogan, line 556: BUCKETSIZE must be an integer between 1 and 255

LiorKogan (approved):
The new limits seem reasonable.
Theoretically, it can be argued that this is a breaking change, but I agree that any command with values beyond these limits is unexpected and most likely wrong.
Also, IIUC, existing filters that are not within these limits (not that we expect any) and not going to be broken, which is good.

YaacovHazan (approved):

tezc:
@LiorKogan yes, I agree it can be considered as breaking change for cf.reserve calls above these limits but as you said, these ones don't work on user passed arguments due to overflow.. This is already buggy.

LiorKogan (approved):

PR #725, opened 2024-01-10
MOD-6344 Fix potential crash for cf.scandump and cf.loadchunk

tezc:
Add boundary checks for cf.scandump and cf.loadchunk.

### DIFF
--- a/src/cf.c
+++ b/src/cf.c
@@ -84,7 +84,7 @@ const char *CF_GetEncodedChunk(const CuckooFilter *cf, long long *pos, size_t *b
 }
 
 int CF_LoadEncodedChunk(const CuckooFilter *cf, long long pos, const char *data, size_t datalen) {
-    if (datalen == 0 || pos < datalen + 1) {
+    if (datalen == 0 || pos <= 0 || (size_t)(pos - 1) < datalen) {
         return REDISMODULE_ERR;
     }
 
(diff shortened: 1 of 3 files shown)


LiorKogan (dismissed):

src/cf.c
```diff
@@ -84,7 +84,7 @@ const char *CF_GetEncodedChunk(const CuckooFilter *cf, long long *pos, size_t *b
 }
 
 int CF_LoadEncodedChunk(const CuckooFilter *cf, long long pos, const char *data, size_t datalen) {
-    if (datalen == 0) {
+    if (datalen == 0 || pos < datalen + 1) {
```
LiorKogan, line 87: Please add protection for datalen = -1

src/cf.c
```diff
@@ -102,6 +102,11 @@ int CF_LoadEncodedChunk(const CuckooFilter *cf, long long pos, const char *data,
         offset -= currentSize;
     }
 
+    // Boundary check before memcpy()
+    if (!filter || filter->bucketSize * filter->numBuckets < offset + datalen) {
```
LiorKogan, line 106: Consider adding checks for protections against overflows

src/cf.c
tezc: It's type `size_t`, cannot be negative.

YaacovHazan (approved):

src/cf.c
LiorKogan: can be max positive (2^32-1) so (2^32-1+1 is 0)

LiorKogan (approved):

PR #744, opened 2024-01-23
Add other platforms in CI.

iddm:
Adds four Ubuntu versions (16.04, 18.04, 20.04, 22.04), Debian Bullseye, RockyLinux 8 and 9, AmazonLinux 2.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,6 @@ on: [push, pull_request]
 jobs:
   test-recent-ubuntu:
     runs-on: ${{ matrix.builder-os }}
-    container:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -178,8 +405,8 @@ jobs:
           submodules: true
       - name: Install Python dependencies
         run: |
+          python3 -m pip install --upgrade pip setuptools wheel
```
tezc, line 408: Do you remember if `wheel` is needed by RLTest as well? Because if I remember correctly, setuptools was needed by RLTest. I guess normally RLTest should declare dependency to these.

iddm: I am not sure either of those is nadeed by RLTest (yet). This here is to solved a Python setup issue on macOS in particular, so not related to the tests and the project.

tezc: Ok, got it. 

Btw, is there anything else you want to add to PR or should we merge it?

iddm: I think we can merge it, but I have no rights to press the "merge" in this project.

tezc (approved):

PR #745, opened 2024-01-23
Add tests for bf and cf, fix findings

tezc:
Added some extra tests for bf and cf
Added fixes/boundary checks to make related parts more robust


src/sb.c
```diff
@@ -26,15 +26,14 @@ bloom_hashval bloom_calc_hash64(const void *buffer, int len);
...
 
 static int SBChain_AddLink(SBChain *chain, uint64_t size, double error_rate) {
-    if (!chain->filters) {
-        chain->filters = RedisModule_Calloc(1, sizeof(*chain->filters));
-    } else {
-        chain->filters =
-            RedisModule_Realloc(chain->filters, sizeof(*chain->filters) * (chain->nfilters + 1));
-    }
+    chain->filters =
+        RedisModule_Realloc(chain->filters, sizeof(*chain->filters) * (chain->nfilters + 1));
```
LiorKogan, line 30: It would change the allocation size when filters=0?
Is it intentional?

tezc: It will still allocate same amount of memory. When `chain->filters` is NULL, `chain->nfilters` is zero. So total allocation is `sizeof(*chain->filters) * 1`.

src/sb.c
```diff
@@ -218,6 +219,25 @@ char *SBChain_GetEncodedHeader(const SBChain *sb, size_t *hdrlen) {
...
+// Returns 0 on success
+int SB_ValidateIntegrity(const SBChain *sb) {
+    if (sb->options &
+        ~(BLOOM_OPT_NOROUND | BLOOM_OPT_ENTS_IS_BITS | BLOOM_OPT_FORCE64 | BLOOM_OPT_NO_SCALING)) {
+        return 1;
+    }
+
+    size_t total = 0;
+    for (size_t i = 0; i < sb->nfilters; i++) {
+        total += sb->filters[i].size;
```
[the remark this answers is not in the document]
tezc: No risk but I added overflow check just to be defensive. Thanks.

YaacovHazan (approved):

PR #750, opened 2024-01-24
Specify redis version in CI.

iddm:
Adds the version specification for Redis in CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,8 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # TODO: figure out the Redis version we need to test on.
-        redis-version: ['unstable']
+        redis-version: ["6.0.20", "7.2.4", "unstable"]
         # jammy and bionic
         builder-os: ['ubuntu-22.04', 'ubuntu-20.04']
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -8,8 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # TODO: figure out the Redis version we need to test on.
-        redis-version: ['unstable']
+        redis-version: ["6.0.20", "7.2.4", "unstable"]
```
tezc, line 11: I feel like a single job with different redis versions is good enough. Others runs might not add extra value. e.g. I don't think we'll ever see an issue that happens on amazon linux+redis 6.0.20 but not on ubuntu+redis 6.0.20 . In my opinion, we can keep this job `test-recent-ubuntu` to run with different redis versions, maybe we can keep `valgrind` job as well. Others might not be necessary. Wdyt?

iddm: It actually happened with centos already.

tezc: ok, let's merge then

tezc (approved):

Issue #753, opened 2024-01-25
Crash on `CMS.MERGE` when weights length does not match numKeys


cunla:
Replicate by running:
```python
def test_cms_merge(r: redis.Redis):
    assert r.cms().initbydim("cmsDim", 100, 5)
    assert r.cms().initbydim("cms2", 100, 5)

    assert r.cms().incrby("cmsDim", ["foo"], [3]) == [3]
    assert r.cms().incrby("cms2", ["foo", "bar"], [4, 1]) == [4, 1]
    assert r.cms().merge("cmsDim", 1, ["cms2"])
    assert r.cms().query("cmsDim", "foo", "bar") == [4, 1]

    with pytest.raises(redis.exceptions.ResponseError, match="CMS: key does not exist"):
        r.cms().merge("noexist", 1, ["cms2"])

    with pytest.raises(redis.exceptions.ResponseError, match="CMS: wrong number of keys"):
        r.cms().merge("cms2", 0, ["cmsDim"])

    with pytest.raises(redis.exceptions.ResponseError, match="wrong number of arguments for 'cms.merge' command"):
        r.cms().merge("cms2", 1, [])

    with pytest.raises(redis.exceptions.ResponseError, match="CMS: wrong number of keys/weights"):
        r.cms().merge("cmsDim", 1, ["cms2", "cms1"], [4, 3])

    with pytest.raises(redis.exceptions.ResponseError, match="CMS: key does not exist"):
        r.cms().merge("cmsDim", 2, ["cms2", "cms1"], [4])

```

```
redis7-1  | === REDIS BUG REPORT START: Cut & paste starting from here ===
redis7-1  | 8:M 25 Jan 2024 14:18:59.714 # Redis 7.2.3 crashed by signal: 11, si_code: 1
redis7-1  | 8:M 25 Jan 2024 14:18:59.714 # Accessing address: 0x79
redis7-1  | 8:M 25 Jan 2024 14:18:59.714 # Crashed running the instruction at: 0xaaaac4b8aea0
redis7-1  |
redis7-1  | ------ STACK TRACE ------
redis7-1  | EIP:
redis7-1  | /opt/redis-stack/bin/redis-server *:6379(RM_StringToLongLong+0x0)[0xaaaac4b8aea0]
redis7-1  |
redis7-1  | Backtrace:
redis7-1  | linux-vdso.so.1(__kernel_rt_sigreturn+0x0)[0xffff97b367a0]
[...]

PR #758, opened 2024-02-18
Fix 753

ashtul:
Fixes issue #753.
In line 193 there was '&&' instead of '||'.
Some streamline of code.


tezc (approved):

tezc:
@ashtul Thank you very much

PR #759, opened 2024-02-20
Delete readies usage from ci-full.yml

tezc:


iddm (approved):
Good, if it works!

tezc:
It works! :)

PR #767, opened 2024-05-14
Add compatible_redis_version field to ramp.yml

tezc:


ramp.yml
```diff
@@ -6,6 +6,7 @@ description: Probabilistic Data Structures for Redis
 homepage: http://redisbloom.io
 license: Redis Source Available License 2.0 (RSALv2) or the Server Side Public License v1 (SSPLv1)
 command_line_args: ""
+compatible_redis_version: "7.4"
```
tezc, line 9:
```suggestion
compatible_redis_version: "7.2"
```

PR #818, opened 2024-11-14
MOD-7497 Add ACL support

ephraimfeldblum:
[MOD-7497](https://redislabs.atlassian.net/browse/MOD-7497)

[MOD-7497]: https://redislabs.atlassian.net/browse/MOD-7497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/common.h
```diff
@@ -6,12 +6,76 @@
...
 #pragma once
 
+#include <stdarg.h>
+#include <string.h>
 #include "redismodule.h"
 #if defined(DEBUG) || !defined(NDEBUG)
 #include "readies/cetara/diag/gdb.h"
 #endif
 
+#define rm_malloc RedisModule_Alloc
```
eranhd, line 16: why not use RedisModule_Alloc instead?

eranhd: pleas add a comment

src/common.h
```diff
@@ -6,12 +6,83 @@
...
+        return REDISMODULE_ERR;
+    }
+    if (!acls_empty) {
+        rm_free((void *)categories);
+    }
+    return REDISMODULE_OK;
+}
+
+#define RegisterCommandWithModesAndAcls(ctx, cmd, f, mode, acls)                                   \
+    if (RedisModule_CreateCommand(ctx, cmd, f, mode, 1, 1, 1) != REDISMODULE_OK ||                 \
```
eranhd, line 78: add #error if MODULE_ACL_CATEGORY_NAME doesn't exists

src/rm_topk.c
```diff
@@ -332,25 +332,32 @@ static size_t TopKMemUsage(const void *value) {
...
-    RMUtil_RegisterReadCmd(ctx, "topk.list", TopK_List_Cmd);
-    RMUtil_RegisterReadCmd(ctx, "topk.info", TopK_Info_Cmd);
+#define MODULE_ACL_CATEGORY_NAME "topk"
+    RegisterAclCategory(ctx);
+    RegisterCommandWithModesAndAcls(ctx, "topk.reserve", TopK_Create_Cmd, "write deny-oom",
+                                    "write fast");
+    RegisterCommandWithModesAndAcls(ctx, "topk.add", TopK_Add_Cmd, "write deny-oom", "write");
+    RegisterCommandWithModesAndAcls(ctx, "topk.incrby", TopK_Incrby_Cmd, "write deny-oom", "write");
+    RegisterCommandWithModesAndAcls(ctx, "topk.query", TopK_Query_Cmd, "readonly", "read");
+    RegisterCommandWithModesAndAcls(ctx, "topk.count", TopK_Count_Cmd, "write", "write");
```
eranhd, line 357: check if should be read

src/common.h
ephraimfeldblum: thinking more on this, i don't think we can #error here because the preprocessor doesn't expand the macro. i think your original idea of asserting here is necessary

ephraimfeldblum: thinking even more on this, if the identifier is not defined, this will miscompile regardless. any assert on the token will not even be called because the line calling the assert will miscompile

src/rm_topk.c
ephraimfeldblum: you're right, should be read. `GetTopKKey(ctx, argv[1], &topk, REDISMODULE_READ)`

src/common.h
```diff
@@ -6,12 +6,83 @@
...
+    return res;
+}
+
+static inline int rm_asprintf(char **restrict str, char const *restrict fmt, ...) {
+    va_list args;
+    va_start(args, fmt);
+    int res = rm_vasprintf(str, fmt, args);
+    va_end(args);
+    return res;
+}
```
iddm, line 49: So you also had to implement it yourself :-)

src/common.h
iddm: I didn't even understand which error @eranhd wanted exactly, so I am just leaving it as-is.

iddm: Ah, I see now. Why don't we pass it here as a parameter? This will be lengthy - yes. But this will also make macro pure in the functional meaning, which doesn't depend on any environment and the outcome of which depends solely on the arguments. I see the code further constantly redefining a name and all this dependency on the environment... I don't like, but do as you wish, I guess. But then, the preprocessor should be able to see the string literal if we pass it as a string literal, shouldn't it? And the defined strings can be checked at a preprocessing time as Eran suggested.

src/rm_cms.c
```diff
@@ -317,24 +317,32 @@ size_t CMSMemUsage(const void *value) {
...
+        .rdb_save = CMSRdbSave,
+        .aof_rewrite = RMUtil_DefaultAofRewrite,
+        .mem_usage = CMSMemUsage,
+        .free = CMSFree,
+        .defrag = CMSDefrag,
+    };
 
     CMSketchType = RedisModule_CreateDataType(ctx, "CMSk-TYPE", CMS_ENC_VER, &tm);
-    if (CMSketchType == NULL)
+    if (CMSketchType == NULL) {
```
iddm, line 331: :+1:

src/rm_tdigest.c
```diff
@@ -915,31 +915,49 @@ size_t TDigestMemUsage(const void *value) {
...
 
     TDigestSketchType = RedisModule_CreateDataType(ctx, "TDIS-TYPE", TDIGEST_ENC_VER, &tm);
-    if (TDigestSketchType == NULL)
+    if (TDigestSketchType == NULL) {
         return REDISMODULE_ERR;
+    }
+
+#define MODULE_ACL_CATEGORY_NAME "tdigest"
+    RegisterAclCategory(ctx);
+    RegisterCommandWithModesAndAcls(ctx, "tdigest.create", TDigestSketch_Create, "write deny-oom",
```
iddm, line 935: I know the ACL category name and the command name aren't the same, but here it all looks like it could have been a string concatenation of `MODULE_ACL_CATEGORY_NAME` + `".create"`. I do not insist on the change, though.

iddm (commented):
Nice work! I approved in general, but a few minor moments I'd still like to have commented on: the error handling of the acl name constant and what you think about the string concatenation. 

And I didn't check for all the commands having the correct ACL permission requirements.

src/rebloom.c
```diff
@@ -1406,76 +1406,94 @@ int RedisModule_OnLoad(RedisModuleCtx *ctx, RedisModuleString **argv, int argc)
...
-                                               .defrag = BFDefrag};
+    static RedisModuleTypeMethods typeprocs = {
+        .version = REDISMODULE_TYPE_METHOD_VERSION,
+        .rdb_load = BFRdbLoad,
+        .rdb_save = BFRdbSave,
+        .aof_rewrite = BFAofRewrite,
+        .free = BFFree,
+        .mem_usage = BFMemUsage,
+        .defrag = BFDefrag,
+    };
```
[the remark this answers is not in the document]
ephraimfeldblum: i did it. it was bothering me.

src/rm_tdigest.c
ephraimfeldblum: i thought about it, but there's other acl category names where it's not the case. eg bloom != bf

tests/flow/test_acl.py
```diff
@@ -0,0 +1,118 @@
...
+
+class testACL():
+  def __init__(self):
+    self.env = Env(decodeResponses=True)
+  
+  def test_acl_category(self):
+      """Test that the various `bloom` categories was added appropriately in module load"""
+      env = self.env
+      if not server_version_at_least(self.env, '7.4.1'):
+         env.skip()
```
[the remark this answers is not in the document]
ephraimfeldblum: yea, this skip on < 7.4.1 is useless.

unless you think i should conditionally add acl categories depending on if RM_AddACLCategory is defined. but i don't think that would meet the requirements of the task.

src/common.h
ephraimfeldblum: the suggesting was to `#error` if the macro were undefined, but that is given to us for free. we miscompile if the macro is undefined, which is literally the same thing.

iddm: This is fine; at least, this is just one use-case for this macro, and only one "global" is used in this way.
About the `#error`, a clear message instead of being unable to compile something, I think, is a little better. Also, some code users may not understand what's wrong in the future, but if we explicitly tell them via the `#error` macro what is wrong, they will immediately know what to do.

src/rm_tdigest.c
iddm: Okay :(

src/common.h
ephraimfeldblum: unfortunately it is impossible to conditionally apply an `#error` based on specific invocations of a macro. the `#error` is parsed by the preprocessor before the macro is ever invoked, so it will always be invoked because the symbol is not defined when the preprocessor parses the macro definition.

build.log
```diff
@@ -0,0 +1,5 @@
+Compiling src/rebloom.c...
+Compiling src/rm_topk.c...
+Compiling src/rm_tdigest.c...
+Compiling src/rm_cms.c...
+Linking /home/ephraim_feldblum/git/RedisBloom/bin/linux-x64-release/redisbloom.so...
```
iddm, line 5: I am not sure we need this file :-)

PR #825, opened 2024-12-05
MOD-7455 Add configuration options which can change at runtime

ephraimfeldblum:
[MOD-7455](https://redislabs.atlassian.net/browse/MOD-7455)

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -1,0 +1,158 @@
+/*
+ * Copyright Redis Ltd. 2016 - present
+ * Licensed under your choice of the Redis Source Available License 2.0 (RSALv2) or
+ * the Server Side Public License v1 (SSPLv1).
+ */
+
+#include <strings.h>
+#include "config.h"
+
+static int setFloatValue(const char *name, RedisModuleString *value, void *privdata, RedisModuleString **err) {
+    RM_ConfigFloat *config = privdata;
+    double *val = &config->value;
+    double new_val;
+    if (RedisModule_StringToDouble(value, &new_val) != REDISMODULE_OK) {
+        *err = RedisModule_CreateStringPrintf(NULL, "Invalid value for `%s`", name);
+        return REDISMODULE_ERR;
+    }
+    if (new_val < config->min || new_val > config->max) {
+        *err = RedisModule_CreateStringPrintf(
+          NULL, "Value for `%s` must be in the range [%f, %f]", name, config->min, config->max);
+        return REDISMODULE_ERR;
+    }
+    *val = new_val;
+    return REDISMODULE_OK;
+}
+static RedisModuleString *getFloatValue(const char *name, void *privdata) {
+    return RedisModule_CreateStringPrintf(NULL, "%f", ((RM_ConfigFloat *)privdata)->value);
+}
+
+static int setNumericValue(const char *name, RedisModuleString *value, void *privdata, RedisModuleString **err) {
+    RM_ConfigFloat *config = privdata;
+    long long *val = &config->value;
+    long long new_val;
+    if (RedisModule_StringToLongLong(value, &new_val) != REDISMODULE_OK) {
+        *err = RedisModule_CreateStringPrintf(NULL, "Invalid value for `%s`", name);
+        return REDISMODULE_ERR;
+    }
+    if (new_val < config->min || new_val > config->max) {
+        *err = RedisModule_CreateStringPrintf(
+            NULL, "Value for `%s` must be in the range [%lld, %lld]", name, config->min, config->max);
+        return REDISMODULE_ERR;
+    }
+    *val = new_val;
+
+    if (strcasecmp(name, "cf_bucket_size") == 0) {
+        rm_config.cf_initial_size.min = new_val * 2;
+    } else if (strcasecmp(name, "cf_initial_size") == 0) {
+        rm_config.cf_bucket_size.max = min(255, new_val / 2);
+    }
+
+    return REDISMODULE_OK;
+}
+static RedisModuleString *getNumericValue(const char *name, void *privdata) {
+    return RedisModule_CreateStringPrintf(NULL, "%zu", *((size_t *)privdata));
+}
+
+typedef struct {
+    const char *name;
+    const char *defaultValue;
+    unsigned int flags;
+    RedisModuleConfigSetStringFunc setValue;
+    RedisModuleConfigGetStringFunc getValue;
+    RedisModuleConfigApplyFunc applyValue;
+    void *privdata;
+} RM_ConfigVar;
+
+RM_ConfigVar rm_config_vars[] = {
+    {
+        .name = "bf-error-rate",
+        .defaultValue = NULL,
+        .flags = REDISMODULE_CONFIG_DEFAULT,
+        .setValue = setFloatValue,
+        .getValue = getFloatValue,
+        .applyValue = NULL,
+        .privdata = &rm_config.bf_error_rate,
+    },
+    {
+        .name = "bf-initial-size",
+        .defaultValue = NULL,
+        .flags = REDISMODULE_CONFIG_DEFAULT,
+        .setValue = setNumericValue,
+        .getValue = setNumericValue,
+        .applyValue = NULL,
+        .privdata = &rm_config.bf_initial_size,
+    },
+    {
+        .name = "bf-expansion-factor",
+        .defaultValue = NULL,
+        .flags = REDISMODULE_CONFIG_DEFAULT,
+        .setValue = setNumericValue,
+        .getValue = setNumericValue,
+        .applyValue = NULL,
+        .privdata = &rm_config.bf_expansion_factor,
+    },
+    {
+        .name = "cf-bucket-size",
+        .defaultValue = NULL,
+        .flags = REDISMODULE_CONFIG_DEFAULT,
+        .setValue = setNumericValue,
+        .getValue = setNumericValue,
+        .applyValue = NULL,
+        .privdata = &rm_config.cf_bucket_size,
+    },
+    {
+        .name = "cf-initial-size",
+        .defaultValue = NULL,
+        .flags = REDISMODULE_CONFIG_DEFAULT,
+        .setValue = setNumericValue,
+        .getValue = setNumericValue,
+        .applyValue = NULL,
+        .privdata = &rm_config.cf_initial_size,
+    },
+    {
+        .name = "cf-max-iterations",
+        .defaultValue = NULL,
+        .flags = REDISMODULE_CONFIG_DEFAULT,
+        .setValue = setNumericValue,
+        .getValue = setNumericValue,
+        .applyValue = NULL,
+        .privdata = &rm_config.cf_max_iterations,
+    },
+    {
+        .name = "cf-expansion-factor",
+        .defaultValue = NULL,
+        .flags = REDISMODULE_CONFIG_DEFAULT,
+        .setValue = setNumericValue,
+        .getValue = setNumericValue,
+        .applyValue = NULL,
+        .privdata = &rm_config.cf_expansion_factor,
+    },
+    {
+        .name = "cf-max-expansions",
+        .defaultValue = NULL,
+        .flags = REDISMODULE_CONFIG_DEFAULT,
+        .setValue = setNumericValue,
+        .getValue = setNumericValue,
+        .applyValue = NULL,
+        .privdata = &rm_config.cf_max_expansions,
+    },
+};
+
+int RM_RegisterConfigs(RedisModuleCtx *ctx) {
+    RedisModule_Log(ctx, "notice", "Registering configuration options");
+
+    for (int i = 0; i < sizeof rm_config_vars / sizeof *rm_config_vars; ++i) {
+        RM_ConfigVar var = rm_config_vars[i];
+        if (RedisModule_RegisterStringConfig(
+                ctx, var.name, var.defaultValue, var.flags, var.getValue,
+                var.setValue, var.applyValue, var.privdata
+            ) != REDISMODULE_OK
+        ) {
+            RedisModule_Log(ctx, "warning", "Failed to register config option `%s`", var.name);
+            return REDISMODULE_ERR;
+        }
+    }
+
+    return REDISMODULE_OK;
+}
--- a/src/config.h
+++ b/src/config.h
@@ -1,0 +1,99 @@
+/*
+ * Copyright Redis Ltd. 2016 - present
+ * Licensed under your choice of the Redis Source Available License 2.0 (RSALv2) or
+ * the Server Side Public License v1 (SSPLv1).
+ */
+
+#include "redismodule.h"
+
+typedef struct {
+    long long value;
+    long long min;
+    long long max;
+} RM_ConfigNumeric;
+
+typedef struct {
+    double value;
+    double min;
+    double max;
+} RM_ConfigFloat;
+
+/*
+ * RM_Config is a global configuration struct for the module, it can be included from each file,
+ * and is initialized with user config options during module startup
+ */
+typedef struct {
+    /*********************************
+     * BLOOM FILTER CONFIG OPTIONS:  *
+     *********************************/
+    // Error ratio. valid range is [0.0, 1.0]. (default: 0.01)
+    // TODO: consider limiting the range to [0.0, 0.25] as per CONFIG PRD
+    // (https://redislabs.atlassian.net/wiki/spaces/DX/pages/3980198010/PRD+CONFIG+for+modules+arguments)
+    RM_ConfigFloat bf_error_rate;
+    // Initial capacity. valid range is [1, 1e9] (default: 100)
+    RM_ConfigNumeric bf_initial_size;
+    // Expansion factor. valid range is [0, 32768] (default: 2)
+    // 0 is equivalent to NONSCALING
+    RM_ConfigNumeric bf_expansion_factor;
+
+    /*********************************
+     * CUCKOO FILTER CONFIG OPTIONS: *
+     *********************************/
+    // Number of items in each bucket. valid range is [1, 255] (default: 2)
+    RM_ConfigNumeric cf_bucket_size;
+    // Initial capacity. valid range is [2*bucket_size, 1e9] (default: 1024)
+    // TODO: add similar max range validation also to CF.RESERVE's capacity argument.
+    RM_ConfigNumeric cf_initial_size;
+    // Maximum iterations. valid range is [1, 65535] (default: 20)
+    RM_ConfigNumeric cf_max_iterations;
+    // expansion factor. valid range is [0, 32768] (default: 1)
+    // 0 is equivalent to NONSCALING
+    RM_ConfigNumeric cf_expansion_factor;
+    // Maximum expansions. valid range is [1, 65535] (default: 32)
+    RM_ConfigNumeric cf_max_expansions;
+} RM_Config;
+
+RM_Config rm_config = {
+    .bf_error_rate = {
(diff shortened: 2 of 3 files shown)


src/config.c
```diff
@@ -0,0 +1,199 @@
...
+            .value = 32,
+            .min = 1,
+            .max = 65535,
+        },
+};
+
+static int setFloatValue(const char *name, RedisModuleString *value, void *privdata,
+                         RedisModuleString **err) {
+    RM_ConfigFloat *config = privdata;
+    double *val = &config->value;
```
eranhd, line 67: why double and not float?

src/rebloom.c
```diff
@@ -131,40 +126,46 @@ static int BFReserve_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv,
...
-            return RedisModule_ReplyWithError(ctx, "ERR expansion should be greater or equal to 1");
+        if (expansion == 0) {
+            nonScaling = BLOOM_OPT_NO_SCALING;
+        } else if (nonScaling == BLOOM_OPT_NO_SCALING) {
+            return RedisModule_ReplyWithError(ctx, "Nonscaling filters cannot expand");
+        }
+        if (expansion < rm_config.bf_expansion_factor.min ||
+            expansion > rm_config.bf_expansion_factor.max) {
+            return RedisModule_ReplyWithError(ctx,
+                                              "ERR expansion should be in the range [0, 32768]");
```
eranhd, line 168: use the min and max values instead

src/config.c
ephraimfeldblum: It is the original datatype used for `error_rate`, changing the type to `float` would be a breaking change, as it would introduce new inaccuracies that didn't exist before. It is also the datatype expected by Redis's CONFIG command.

And it's not like the size really matters that much. It's a single configuration variable. There's only ever one of them per module. I think we can spare a couple bytes.

src/config.c
```diff
@@ -0,0 +1,199 @@
...
+static RedisModuleString *getNumericValue(const char *name, void *privdata) {
+    return RedisModule_CreateStringPrintf(NULL, "%zu", *((size_t *)privdata));
+}
+
+typedef struct {
+    const char *name;
+    const char *defVal;
+    RedisModuleConfigSetStringFunc setValue;
+    RedisModuleConfigGetStringFunc getValue;
+    void *privdata;
```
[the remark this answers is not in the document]
ephraimfeldblum: Per the CONFIG spec, all configs have `privdata` (private data) that Redis doesn't use. It simply stores it to be passed to the `setValue` and `getValue` functions.

In this case, we use it to store a pointer to the module-side configuration variable so that we can update it whenever `CONFIG SET` is called, and read from it whenever `CONFIG GET` is called. We do not store the config variable on the Redis-side as if we did that then we would need to make an external call every time we wanted to update or read the variable within the module.

src/config.c
```diff
@@ -0,0 +1,197 @@
...
+ * Licensed under your choice of the Redis Source Available License 2.0 (RSALv2) or
+ * the Server Side Public License v1 (SSPLv1).
+ */
+
+#include <strings.h>
+#include "config.h"
+
+// Default configuration values.
+// All values provided by CONFIG PRD
+// (https://redislabs.atlassian.net/wiki/spaces/DX/pages/3980198010/PRD+CONFIG+for+modules+arguments)
```
iddm, line 12: This, probably, shouldn't be here :-)
I understand this would help, but not everyone in the world...

src/config.c
```diff
@@ -0,0 +1,197 @@
...
+
+static int setFloatValue(const char *name, RedisModuleString *value, void *privdata,
+                         RedisModuleString **err) {
+    RM_ConfigFloat *config = privdata;
+    double new_val;
+    if (RedisModule_StringToDouble(value, &new_val) != REDISMODULE_OK) {
+        *err = RedisModule_CreateStringPrintf(NULL, "Invalid value for `%s`", name);
+        return REDISMODULE_ERR;
+    }
+    if (new_val < config->min || new_val > config->max) {
```
iddm, line 72: So the maximum and the minimum values are always inclusive, right? Then `cg_max_expansions` should have `.max = 65536` then, as well as others should be corrected? Do we have a test for the ranges?

src/config.c
```diff
@@ -0,0 +1,197 @@
...
+        return REDISMODULE_ERR;
+    }
+    config->value = new_val;
+    return REDISMODULE_OK;
+}
+static RedisModuleString *getFloatValue(const char *name, void *privdata) {
+    return RedisModule_CreateStringPrintf(NULL, "%f", ((RM_ConfigFloat *)privdata)->value);
+}
+
+static int setNumericValue(const char *name, RedisModuleString *value, void *privdata,
```
iddm, line 84: So we can safely call it `setIntegerValue` then, I guess?

src/config.c
```diff
@@ -0,0 +1,197 @@
...
+    },
+    {
+        .name = "cf-bucket-size",
+        .defVal = "2",
+        .setValue = setNumericValue,
+        .getValue = getNumericValue,
+        .privdata = &rm_config.cf_bucket_size,
+    },
+    {
+        .name = "cf-initial-size",
```
iddm, line 150: Throughout the code, I see the same string literals. It would be helpful to have those extracted into named string constants and used throughout the code instead.

src/config.c
```diff
@@ -0,0 +1,197 @@
...
+    {
+        .name = "cf-expansion-factor",
+        .defVal = "1",
+        .setValue = setNumericValue,
+        .getValue = getNumericValue,
+        .privdata = &rm_config.cf_expansion_factor,
+    },
+    {
+        .name = "cf-max-expansions",
+        .defVal = "32",
```
iddm, line 172: So if we have those defined as strings using the new config API, will the user receive those also as `"32"` in the output of `GET`? This **might** be confusing. Even though I understand the current problem with the API, I just was wondering if there is a way to specifically address this in particular?

src/config.c
```diff
@@ -0,0 +1,197 @@
...
+int RegisterStringConfig(RedisModuleCtx *ctx, RM_ConfigVar var) {
+    return RedisModule_RegisterStringConfig(ctx, var.name, var.defVal,
+                                            REDISMODULE_CONFIG_UNPREFIXED, var.getValue,
+                                            var.setValue, NULL, var.privdata);
+}
+
+int RM_RegisterConfigs(RedisModuleCtx *ctx) {
+    RedisModule_Log(ctx, "notice", "Registering configuration options");
+
+    for (int i = 0; i < sizeof rm_config_vars / sizeof *rm_config_vars; ++i) {
```
iddm, line 188: Can we please use `sizeof` with the function syntax? :-)
Also here - comparing the signed and unsigned integers within the same row. And later, using `i` declared as `int` as an index for the `rm_config_vars` which also requires an unsigned as the receiving type.

src/config.c
```diff
@@ -0,0 +1,197 @@
...
+        RM_ConfigVar var = rm_config_vars[i];
+        if (RegisterStringConfig(ctx, var) != REDISMODULE_OK) {
+            RedisModule_Log(ctx, "warning", "Failed to register config option `%s`", var.name);
+            return REDISMODULE_ERR;
+        }
+    }
+
+    return RedisModule_LoadConfigs(ctx);
```
iddm, line 196: Should this function really have two responsibilities? Why can't we simply rename it then, at the very least, to `RegisterAndLoadConfiguration`?

src/config.h
```diff
@@ -0,0 +1,70 @@
...
+     * BLOOM FILTER CONFIG OPTIONS:  *
+     *********************************/
+    // Error ratio. valid range is [0.0, 1.0]. (default: 0.01)
+    // TODO: consider limiting the range to [0.0, 0.25] as per CONFIG PRD
+    RM_ConfigFloat bf_error_rate;
+    // Initial capacity. valid range is [1, 1e9] (default: 100)
+    RM_ConfigNumeric bf_initial_size;
+    // Expansion factor. valid range is [0, 32768] (default: 2)
+    // 0 is equivalent to NONSCALING
+    RM_ConfigNumeric bf_expansion_factor;
```
iddm, line 38: I believe, there is almost no reason for this comments to exists here, when we already have a well-defined and easily readable table of the default values and the ranges for each of these.  I'd delete the comments explaining the valid ranges and so on, and leave only the comments explaining what it is etc.

tests/flow/test_overall.py
```diff
@@ -196,7 +196,7 @@ def test_insert(self):
         with env.assertResponseError():
             env.cmd('bf.insert', 'missingFilter', 'ERROR', 'big')
         with env.assertResponseError():
-            env.cmd('bf.insert', 'missingFilter', 'EXPANSION', '0', 'ITEMS', 'foo')
+            env.cmd('bf.insert', 'missingFilter', 'EXPANSION', '-1', 'ITEMS', 'foo')
```
iddm, line 199: I am curious what exactly required this change when we needed only to change the configuration options-related code?

tests/flow/test_overall.py
```diff
@@ -411,10 +411,10 @@ def test_invalid_expansion(self):
         env = self.env
         env.cmd('FLUSHALL')
         try:
-            env.cmd('BF.RESERVE b 0.1 1 EXPANSION 0')
+            env.cmd('BF.RESERVE b 0.1 1 EXPANSION -1')
             env.assertTrue(False)
         except Exception as e:
-            env.assertEqual(str(e), 'expansion should be greater or equal to 1')
+            env.assertEqual(str(e), 'expansion must be in the range [0, 32768]')
```
iddm, line 417: Can we instead `assert "expansion must be in the range"  in str(e)`?

src/config.c
ephraimfeldblum: the max and min are always inclusive. thanks for noticing this one. there are no others that needed correcting. tests will be coming.

tests/flow/test_overall.py
ephraimfeldblum: 0 expansions is (now) valid. and equivalent to NONSCALING

src/config.c
ephraimfeldblum: sizeof with function syntax on an expression is :vomiting_face:

ephraimfeldblum: arrays may be indexed by `int`s in C. an index array does not need to be an unsigned type in C. because C is special like that.

src/config.c
ephraimfeldblum: Even if we did solve this in the specific case, I don't really see a way around it in the general case. It's not any different than `CONFIG GET bf-error-rate` returning `"0.01"` and that has no workaround.

src/config.c
iddm: Have I ever told you I hated C for stupid things like that? I wonder how they would address (and call it perfectly valid) an array with an index of `-100`.

ephraimfeldblum: `arr[idx]` is simply sugar for `*(arr + idx)`. subtraction of an integral value from a pointer is valid in C as long as the resulting pointer obtained from the subtraction is a valid pointer to within the original array object.
hope that helps :sweat_smile:

iddm: > `arr[idx]` is simply sugar for `*(arr + idx)`. subtraction of an integral value from a pointer is valid in C as long as the resulting pointer obtained from the subtraction is a valid pointer to within the original array object. hope that helps 😅

Yes, I read the K&R book a long time ago. This doesn't make C better. As well as iterating over an array till its end with a signed integer :-)

src/config.c
```diff
@@ -0,0 +1,149 @@
...
+                                              name, config->min, config->max);
+        return REDISMODULE_ERR;
+    }
+    config->value = new_val;
+    RedisModule_FreeString(NULL, config->str_value);
+    config->str_value = RedisModule_HoldString(NULL, value);
+    return REDISMODULE_OK;
+}
+static RedisModuleString *getFloatValue(const char *name, void *privdata) {
+    return ((RM_ConfigFloat *)privdata)->str_value;
```
iddm, line 88: Ah. Nevermind.

src/config.c
```diff
@@ -0,0 +1,149 @@
...
+        return REDISMODULE_ERR;
+    }
+    config->value = new_val;
+    RedisModule_FreeString(NULL, config->str_value);
+    config->str_value = RedisModule_HoldString(NULL, value);
+
+    if (!RM_ConfigStrCaseCmp(name, cf_bucket_size)) {
+        rm_config.cf_initial_size.min = new_val * 2;
+    } else if (!RM_ConfigStrCaseCmp(name, cf_initial_size)) {
+        rm_config.cf_bucket_size.max = new_val / 2 < 255 ? new_val / 2 : 255;
```
iddm, line 112: :-D

Please, wrap the comparing expressions in `()`, just for clarity.

src/config.c
```diff
@@ -0,0 +1,149 @@
...
+            RedisModule_CreateStringPrintf(NULL, "Value for `%s` must be in the range [%lld, %lld]",
+                                           name, config->min, config->max);
+        return REDISMODULE_ERR;
+    }
+    config->value = new_val;
+    RedisModule_FreeString(NULL, config->str_value);
+    config->str_value = RedisModule_HoldString(NULL, value);
+
+    if (!RM_ConfigStrCaseCmp(name, cf_bucket_size)) {
+        rm_config.cf_initial_size.min = new_val * 2;
```
iddm, line 110: If the `new_val` is > `127`, the multiplication will make this higher than the defined maximum value for the configuration, I just wanted to check if this is correct for the `cf_initial_size` and the way it is used.

src/config.c
```diff
@@ -0,0 +1,149 @@
...
+        const char *default_val = RedisModule_StringPtrLen(rm_config.config.str_value, NULL);      \
+        if (RedisModule_RegisterStringConfig(ctx, name, default_val,                               \
+                                             REDISMODULE_CONFIG_UNPREFIXED,                        \
+                                             _Generic(rm_config.config.value,                      \
+                                             long long: getIntegerValue,                           \
+                                             double: getFloatValue),                               \
+                                             _Generic(rm_config.config.value,                      \
+                                             long long: setIntegerValue,                           \
+                                             double: setFloatValue),                               \
+                                             NULL, &rm_config.config) != REDISMODULE_OK) {         \
```
iddm: Can we have a static function for this code?

```c
        const char *name = RM_ConfigOptionToString(config);                                        \
        rm_config.config.str_value = RedisModule_CreateStringPrintf(                               \
            NULL, _Generic(rm_config.config.value, long long: "%lld", double: "%f"),               \
            rm_config.config.value);                                                               \
        const char *default_val = RedisModule_StringPtrLen(rm_config.config.str_value, NULL);      \
        if (RedisModule_RegisterStringConfig(ctx, name, default_val,                               \
                                             REDISMODULE_CONFIG_UNPREFIXED,                        \
                                             _Generic(rm_config.config.value,                      \
                                             long long: getIntegerValue,                           \
                                             double: getFloatValue),                               \
                                             _Generic(rm_config.config.value,                      \
                                             long long: setIntegerValue,                           \
                                             double: setFloatValue),                               \
                                             NULL, &rm_config.config) != REDISMODULE_OK) { 
```

And within this macro, just invoke it within the same `do while(0)` loop. In my opinion, this would make the code a little more readable and concise. This function, though, would be better to have the `rm_config` passed as an argument as well, to make it purer, but this shouldn't be a problem to do from within this macro.

iddm (commented):
I am a little confused, so the CF and BF settings are parsed differently?

src/config.c
ephraimfeldblum: `cf-initial-size`'s minimum may be greater than 255, that's fine. its minimum is `max(0, cf-bucket-size * 2)`, but since `cf-bucket-size` is bounded by `[1 .. 255]`, which is strictly greater than 0, we don't need to clamp.

src/config.c
```diff
@@ -0,0 +1,149 @@
+/*
```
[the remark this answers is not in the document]
ephraimfeldblum: there was no old way of setting configuration options at runtime. default config variables could be passed in module args at module load, and we still support the old names (with a deprecation log message).

individual filter configurations (such as `CF.RESERVE <KEY> <CAPACITY> [BUCKETSIZE] [MAXITERATIONS] [EXPANSION]`) are untouched. their names remain as before, they are not modified at all.

src/rebloom.c
```diff
@@ -525,46 +534,52 @@ static int CFReserve_RedisCommand(RedisModuleCtx *ctx, RedisModuleString **argv,
...
     if (mi_loc != -1) {
         if (RedisModule_StringToLongLong(argv[mi_loc + 1], &maxIterations) != REDISMODULE_OK) {
             return RedisModule_ReplyWithError(ctx, "Couldn't parse MAXITERATIONS");
-        } else if (maxIterations <= 0 || maxIterations > CF_MAX_ITERATIONS) {
-            return RedisModule_ReplyWithError(
-                ctx, "MAXITERATIONS: value must be an integer between 1 and 65535, inclusive.");
+        }
+        if (!isConfigValid(maxIterations, rm_config.cf_max_iterations)) {
+            return RedisModule_ReplyWithErrorFormat(
+                ctx, "MAXITERATIONS: value must be in the range [%lld, %lld]",
```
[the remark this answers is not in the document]
ephraimfeldblum: `MAXITERATIONS` is not a configuration option. it is a parameter to `CF.RESERVE` which is used to override the configuration option for the specific filter being reserved.

src/rebloom.c
```diff
@@ -1333,20 +1356,60 @@ static int rsStrcasecmp(const RedisModuleString *rs1, const char *s2) {
...
+        } else {                                                                                   \
+            BAIL(configRangeFormat(config), RM_ConfigOptionToString(config), rm_config.config.min, \
+                 rm_config.config.max);                                                            \
+        }                                                                                          \
     } while (0)
 
+#define tryGetConfigFromArgs(ctx, argv, idx, configNameLegacy, configName)                         \
+    ({                                                                                             \
+        bool legacy = false;                                                                       \
+        if ((!rsStrcasecmp(argv[idx], configNameLegacy) && (legacy = true)) ||                     \
```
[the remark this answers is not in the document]
ephraimfeldblum: yes. i couldn't figure out a way of doing this without either the assignment within the condition or splitting this up into 2 distinct if statements with duplicate code between them, ie
```
if (!rsStrcasecmp(argv[idx], configNameLegacy)) {
  getConfigFromString(argv[idx + 1], configName);
  RedisModule_Log(ctx, "warning",
    "The '" configNameLegacy "' configuration is deprecated. Please use '%s' instead",
    RM_ConfigOptionToString(configName));
  continue;
}
if (!RM_ConfigRMStrCaseCmp(argv[idx], configName)) {
  getConfigFromString(argv[idx + 1], configName);
  continue;
}
```

src/config.c
iddm: You can still extract this info a function and have the if and the return statements just as you described.

src/rebloom.c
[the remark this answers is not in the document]
iddm: I apologise :-)

src/rebloom.c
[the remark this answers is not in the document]
iddm: How about 

```c
const bool isLegacy = !rsStrcasecmp(argv[idx], configNameLegacy);
```

and then just as it was?

src/config.c
ephraimfeldblum: unfortunately not. because there is a lot of token pasting in the macro, which cannot be done at runtime. `rm_config.config.value` expands to `rm_config.<actual-config-variable-name>.value`, etc

iddm (approved):

eranhd (approved):

iddm (approved):

PR #827, opened 2024-12-09
Fix MOD-8194, initialize lookup table on rdbload.

MeirShpilraien:


src/rm_topk.c
```diff
@@ -308,6 +309,11 @@ static void *TopKRdbLoad(RedisModuleIO *io, int encver) {
         }
     }
 
+    /* Initialize lookupTable */
+    for (uint32_t i = 0; i < TOPK_DECAY_LOOKUP_TABLE; ++i) {
+        topk->lookupTable[i] = pow(topk->decay, i);
```
eranhd, line 314: @MeirShpilraien can we build a shared lookup table mechanism to reduce the memory overhead here?

eranhd (approved):

eranhd (commented):
you tests fails on valgrind

src/rm_topk.c
MeirShpilraien: We probably can but it is out of the scope of this fix, the goal here is to fix a bug :)

eranhd (approved):

PR #901, opened 2025-09-17
MOD-11490  upload Bloom beta version to S3

gabsow:


.github/actions/setup-env/action.yml
```diff
@@ -9,6 +9,9 @@ inputs:
   redis-ref:
     description: Redis ref
     required: false
+  nightly-version:
```
eranhd, line 12: on JSON you called it beta-version

gabsow: fixed

eranhd (approved):

PR #1008, opened 2026-05-17
MOD-15731 - modules modernization 

TalBarYakar:
> [!NOTE]
> **Medium Risk**
> Wide CI/Docker/bootstrap changes across many OS images can break builds or tests if package sets or venv activation diverge from runners; run-tests now fails workflows on test errors instead of masking them.
> 
> **Overview**
> This PR **modernizes build/CI bootstrap** around a single **OSNICK-based** `.install` flow: `install_script.sh` detects the distro, installs packages via shared `lib/pm.sh` / `lib/packages.sh` and per-OS scripts under `.install/os/`, then provisions **`uv` + repo `venv` + pip deps** in `lib/setup-python.sh`. Legacy per-distro install snippets and `common_installations.sh` are removed in favor of that stack.
> 
> **`make bootstrap`** is added at the top of the Makefile so hosts can install deps **without parsing Readies** first; help text shifts from `setup` to `bootstrap`. **Dockerfiles** largely drop inline `apt`/`dnf` package lists and call `install_script.sh` instead (Alpine keeps a special system-venv path).
> 
> **GitHub Actions** gain optional **`use-venv`**: a new **`install-python-deps`** action activates the bootstrap venv (with deduped `bash_profile` lines) and only runs extras/`sbin/setup` when asked; build/pack/test actions optionally activate `venv` and enable **devtoolset-11** where needed. **`run-tests`** no longer creates a venv or calls the old common install script; it **fails the job on test failure** (`PIPESTATUS`) instead of swallowing exit codes with `|| true`. **macOS flow** wires these actions end-to-end after `.install` setup.
> 
> **Flow tests** pick an RLTest-capable Python (`venv` first, then 3.12→3.9) via **`RLTEST_PYTHON`**. **`module.conf`** adds **redis-gen-conf private** markers only (no new tunables exported).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 17a38055a501430bb136be3172359f14f02627f9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


gabsow (changes requested):
## Code Review — MOD-15731 modules modernization

Overall this is a solid refactor, but there are several issues ranging from blockers to nits that should be addressed before merging.

---

### BLOCKERS

**1. `_pm_enable_el8_extras` failure not propagated** (`.install/lib/pm.sh`)
The function correctly returns 1 on failure, but the caller does not check:
\`\`\`bash
_pm_enable_el8_extras   # failure silently ignored
rhel_default_install    # runs on broken system
\`\`\`
Fix: \`_pm_enable_el8_extras || exit 1\`

**2. Alpine Dockerfile — venv creation unguarded** (\`Dockerfile.alpine\`)
\`\`\`dockerfile
RUN python3 -m venv --system-site-packages /opt/.venv
\`\`\`
No error handling. If this fails, the next line sets \`ENV PATH="/opt/.venv/bin:..."\` and subsequent \`pip install\` commands fail with a confusing error. The new \`install_script.sh\` flow checks \`[ ! -x "$ROOT/venv/bin/python" ]\` and exits cleanly — Alpine should do the same.

**3. \`setup-python.sh\` returns 0 on uv failure**
\`\`\`bash
if ! command -v uv >/dev/null 2>&1; then
    echo "WARNING: uv installation failed; skipping venv setup" >&2
    return 0   # swallows the failure
fi
\`\`\`
The caller continues without a venv. Change to \`exit 1\` so the failure is explicit.

**4. Venv race condition** (`.install/lib/setup-python.sh`)
The check-then-delete pattern:
\`\`\`bash
if [ -d "$ROOT/venv" ] && [ ! -x "$ROOT/venv/bin/python" ]; then
    rm -rf "$ROOT/venv"
fi
\`\`\`
is not atomic. Two parallel CI runners on the same machine can race. Add a lockfile or document that bootstrap must not run in parallel.

---

### SHOULD-FIX

**5. Alpine — does not verify that \`uv\` is absent for all subsequent steps**
Old Dockerfile installed \`uv\` explicitly; new one skips it with a comment about clang flags. If any later step (or a developer running \`make bootstrap\`) assumes \`uv\` is in PATH on Alpine, it will fail silently. Audit all Alpine \`RUN\` steps and composite actions for \`uv\` usage.

**6. \`make test VG="" SAN=""\` — verify empty strings disable correctly**
The new \`run-tests\` action maps inputs to env vars and passes \`VG="$INPUT_RUN_VALGRIND"\`. Confirm the Makefile treats an empty string as disabled — if it checks \`ifdef VG\` instead of \`ifneq ($(VG),)\`, the behavior is wrong.

**7. \`SETUP_PYTHON_VERSION\` is an undocumented side-effect**
\`pm.sh\` exports \`SETUP_PYTHON_VERSION=3.11\` for EL8, but \`setup-python.sh\` doesn't document this dependency. A developer running \`setup-python.sh\` directly gets Python 3.12 silently. Add a comment at the top of \`setup-python.sh\` explaining the variable.

**8. Shared apt update cache across \`apt_install\` calls**
\`pm.sh\` sets \`_pm_apt_updated=1\` after the first \`apt-get update\` and skips it on subsequent calls. If any OS file interleaves \`apt_install\` with external package manager commands that add new repos, the cache will be stale. Verify all OS files do package installs in a single logical group.

**9. \`Makefile bootstrap\` suppresses all output with \`@\`**
\`\`\`makefile
bootstrap:
    @rm -rf $(ROOT)/venv
    @cd $(ROOT)/.install && ./install_script.sh $(INSTALL_SCRIPT_MODE)
\`\`\`
When \`install_script.sh\` fails midway, there's no indication of which command failed. Consider removing \`@\` or adding \`set -x\` for debuggability.

---

### NITS

**10. \`src/version.h\` hardcodes \`8.7.90\`** — the comment says this is overridden with \`-D\` at compile time, but the diff doesn't show where those flags are set in CI. Confirm this is wired up correctly so the wrong version isn't advertised.

**11. \`module.conf\` \`:private\` markers** — verify that \`sync-redis-conf.mk\` (not shown in diff) correctly parses the \`BEGIN redis-gen-conf:private\` / \`END redis-gen-conf:private\` boundaries. If not, these are dead comments.

**12. \`common_installations.sh\` removal — \`tests/flow/requirements.txt\` now conditional**
Old script assumed the file existed; new \`setup-python.sh\` skips it if absent. Make sure all checkout configurations (shallow, sparse) that run tests include this file, or make its absence a hard error.

gabsow:
**Version mismatch in `src/version.h` (and likely `pack/ramp.yml`)**

This PR targets `master` but sets the version to `8.7.90` — that's incorrect.

Expected versions:
- `master` → `8.9.70`
- `8.8` branch → `8.8.0` (already correct on that branch)

The `8.8` branch already has the right `8.8.0` in `version.h`. Please update `master` to `8.9.70` and check `pack/ramp.yml` for the same issue.

gabsow:
**`pack/ramp.yml` — wrong versions for master**

The PR changes:
```yaml
-compatible_redis_version: "99.99"
-min_redis_version: "7.4"
+compatible_redis_version: "8.8"
+min_redis_version: "8.8"
```

On `master` these should stay as:
```yaml
compatible_redis_version: "99.99"
min_redis_version: "7.4"
```

The `8.8` values belong on the `8.8` branch only. Please revert these two lines.

TalBarYakar:
remark 1- Fixed. Changed line 168 from _pm_enable_el8_extras to _pm_enable_el8_extras || exit 1 and added a comment above it explaining the failure-mode rationale (without codeready-builder, the follow-up dnf_install of libatomic-devel resolves to nothing and the user gets a "no package" error several steps removed from the actual cause — propagating the existing non-zero return code surfaces the original "Fix RHUI/subscription repos" message from _pm_enable_el8_extras at the right spot). One-line change + 3 lines of comment, no logic added.

remark 2 - Fixed. Chained && test -x /opt/.venv/bin/python into the same RUN so the layer aborts if the venv's interpreter isn't usable.

Worth noting on the framing: a true non-zero exit from python3 -m venv already aborts the Docker build at this line (the ENV lines wouldn't run, so the "pip install fails confusingly" path doesn't apply for hard failures). The case the new check actually closes is the silent-partial-success one — venv exits 0 but produces a directory without a working bin/python (broken ensurepip, dangling symlink). Without the guard, downstream pip install would then fail several RUNs later with a misleading "No module named pip" / "bad interpreter" error. The added test -x mirrors the same defensive check install_script.sh's setup-python.sh runs on the native bootstrap path, so the two flows match.

remark 3- Fixed. Changed return 0 2>/dev/null || exit 0 to return 1 2>/dev/null || exit 1, updated the message from WARNING: ... skipping venv setup to ERROR: ... cannot create venv, and added a short two-line comment explaining why this must be a hard failure: a “successful” bootstrap without a venv on disk silently breaks every downstream test/flow step, and this invariant should fail loudly at the source rather than waiting for someone to notice it in CI half an hour later. Two-line behavior change plus a one-line explanatory comment.

remark 4- Not changed — the race only exists if two bootstraps are deliberately run concurrently against the same checkout (not a supported mode), while our CI/workspace model uses isolated per-job directories, so adding lockfile/trap complexity here felt like over-engineering for a scenario we don’t currently have.


remark 5- Audited — Alpine still installs and uses uv correctly via install_script.sh → setup-python.sh; no hidden uv consumers were found, and the only notable issue is that the image currently ends up with two venvs ($ROOT/venv and /opt/.venv), with the former effectively unused at runtime.


remark 6 -Every VG/SAN check across the repo uses the value-based form, not ifdef:

Makefile:137 — ifeq ($(VG),1)
Makefile:211 — ifneq ($(SAN),)
deps/readies/mk/valgrind.cfg and clang-sanitizer.{cfg,defs} — same patterns
Grepped the whole tree (Makefile, tests, readies framework) for \bifdef\s+(VG|SAN) — zero hits.

Cross-checking with the action input defaults: run-tests/action.yml sets run_valgrind to '0' and run_sanitizer to '', then passes them as VG="$INPUT_RUN_VALGRIND" SAN="$INPUT_RUN_SANITIZER". So a no-args invocation lands as make test VG=0 SAN= — ifeq (0,1) is false (valgrind off) and ifneq (,) is false (sanitizer off). Both disable as intended.

remark 7 - fixed. Added a 7-line "Optional input: SETUP_PYTHON_VERSION" block right after the existing "Replaces the legacy …" paragraph in the script header. It calls out: the default (3.12), who currently overrides it (pm.sh's el8_default_install), and why (EL8 base python3 is 3.6, and uv's auto-3.12 breaks psutil's aarch64 source build). Anyone reading setup-python.sh standalone now sees the lever and the reason without having to grep pm.sh. Pure documentation change — no behavior touched.

remark 8- no current OS file breaks the cache contract; bionic.sh is the only interleaving case, and add-apt-repository -y already refreshes apt indexes implicitly, so no stale-cache bug exists today (though adding an explicit cache-invalidate helper later could harden against future changes).

regarding the version lets talk about it tommorow i am not sure what should be done

TalBarYakar:
change the version.h file to 99.99.99

gabsow:
Verified each remark against latest HEAD (`17a3805`):

**Blockers 1-3:** ✅ confirmed fixed.

**Blocker 4 (venv race):** ack — your CI-isolation argument is fair; accepting as won't-fix.

**Should-fix 5-8:** verified per your audit.
- 5: `make bootstrap` on Alpine does install uv via `setup-python.sh`; the dual-venv situation (`$ROOT/venv` + `/opt/.venv`) is worth a follow-up but doesn't block.
- 6: confirmed `ifeq ($(VG),1)` + `ifneq ($(SAN),)`; `VG=0 SAN=` disables both correctly.
- 7: the new doc block at `setup-python.sh:11-16` reads well.
- 8: agreed — `add-apt-repository -y` refreshes implicitly so bionic is fine today.

**Nits:**
- 9 (Makefile `@`): ✅ `install_script.sh` line no longer silenced.
- 10 (`version.h`): ✅ reverted to `99.99.99`.
- 11 (`module.conf` markers): ⚠️ **Cursor Bugbot caught a delimiter mismatch I missed** — END opens with `<<<` instead of `>>>`:
  ```
  # >>> BEGIN redis-gen-conf:private <<<
  # <<< END redis-gen-conf:private <<<
  ```
  If the parser in `modules/sync-redis-conf.mk` pairs `>>> BEGIN <<< … >>> END <<<`, parsing will break. Quick fix: `# <<< END redis-gen-conf:private <<<` → `# >>> END redis-gen-conf:private <<<`.
- 12 (`tests/flow/requirements.txt`): ✅ fine — Dockerfiles use `requirements_docker.txt` by design.

**One additional finding worth a quick look:**
`.github/actions/pack-module/action.yml:29-30` still interpolates `${{ inputs.beta-version }}` directly inside the `run:` block — same script-injection class SonarCloud flagged for the other inputs (which you fixed via `INPUT_*` env vars). Same pattern would resolve it:
```yaml
env:
  INPUT_BETA_VERSION: ${{ inputs.beta-version }}
run: |
  if [[ -n "$INPUT_BETA_VERSION" ]]; then
    export BETA_VERSION="$INPUT_BETA_VERSION"
  fi
```

Once the `module.conf` END marker is fixed (and ideally the `pack-module` env-var change), I'll dismiss the CHANGES_REQUESTED.

Re: `ramp.yml` `compatible_redis_version` — happy to discuss tomorrow.

.install/lib/setup-python.sh
```diff
@@ -0,0 +1,68 @@
...
+fi
+
+# A stale or partial venv (e.g. a previous `make bootstrap` aborted halfway,
+# or the developer ran `python3 -m venv` against a now-missing python) shows
+# up as `$ROOT/venv` existing but `bin/python` not being executable. Wipe
+# and recreate so we don't trip the executable check below.
+if [ -d "$ROOT/venv" ] && [ ! -x "$ROOT/venv/bin/python" ]; then
+    echo "==> [redisbloom] $ROOT/venv looks broken (no bin/python); recreating"
+    rm -rf "$ROOT/venv"
+fi
```
gabsow, line 43: **Re: blocker #4 (venv check-then-delete race) — suggested resolution: won't-fix + a one-line comment, not a lockfile.**

A `flock`/`mkdir` lock would close the race but introduces its own failure modes (stale locks if a process dies mid-bootstrap, sleep-loop on a wedged lock, `EXIT` trap correctness) for a scenario that doesn't actually occur: CI uses isolated per-job workspaces, and local `make bootstrap` is run sequentially.

Proposed: drop a comment above this block documenting the constraint so the assumption is captured in code, e.g.:

```bash
# Not safe for concurrent invocation against the same $ROOT — caller must
# serialize. CI runs each job in an isolated workspace and humans don't
# parallelize `make bootstrap`, so we don't take a lock here.
```

With that added, I'm fine treating this remark as resolved.

gabsow (approved):
Approved — all 4 blockers resolved or accepted as won't-fix per discussion above:

- Blockers 1-3: fixed in latest HEAD.
- Blocker 4 (venv race): accepted as won't-fix; will be documented in-code per [inline suggestion](https://github.com/RedisBloom/RedisBloom/pull/1008#discussion_r3301677011).

Remaining nice-to-haves (non-blocking, can ship and follow up):
- `module.conf:10` END marker delimiter (`<<<` → `>>>`) — Bugbot finding; verify the `sync-redis-conf.mk` parser before fixing.
- `pack-module/action.yml:29-30` script-injection pattern — move `inputs.beta-version` to an `INPUT_*` env var like the other actions.

`ramp.yml` `compatible_redis_version` to discuss tomorrow as planned.

PR #1028, opened 2026-06-29
MOD-16514 - fix the bootstrap and build

TalBarYakar:
> [!NOTE]
> **Medium Risk**
> Changes affect multi-OS bootstrap scripts and compiler/python alternatives used by all CI builds; mistakes could break builds on specific distros but don’t touch runtime Redis module logic.
> 
> **Overview**
> Hardens **CI/bootstrap** across distros so builds get a consistent compiler, Python, and AWS CLI without fighting shared containers or apt’s clang defaults.
> 
> **Shared AWS CLI v2** is added via `install_aws_cli()` (arch-aware, no-op if `aws` exists) and replaces duplicated installers on Rocky/Mariner/Amazon Linux and pip-based install on Azure Linux 3.
> 
> **Toolchain pinning:** `debian_default_install` forces `cc`/`gcc`/`g++` to the newest system GCC so clang doesn’t win after `clang` packages install. EL8/EL9 (and Amazon Linux 2 devtoolset) symlink gcc-toolset binaries into `/usr/local/bin`; EL8 also installs `python3.11-pip` and pins `python3` to 3.11. Ubuntu **bionic** gets noninteractive apt, manual toolchain PPA setup, **only upgrades** to gcc-10 when the active compiler is older than 10, and builds cmake 3.28 only when needed; **focal** uses the same conditional gcc-10 logic.
> 
> **Flow tests:** `RLTest` is bumped to 0.7.27 with `numpy<2`; a cuckoo overflow test tolerates `ResponseError` on a huge `CF.INSERT`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4b68c79990997a91a97ce78f57631b2e6150d179. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.install/lib/packages.sh
+++ b/.install/lib/packages.sh
@@ -8,6 +8,7 @@
 # with the system interpreter (see tests/flow/requirements.txt).
 #
 # Sourced by os/<osnick>.sh after lib/pm.sh. All variables here are plain
+# space-separated strings so callers can splat them with `apt_install $SET`.
 
 # Install AWS CLI v2 from the official installer (arch-aware). Skips if
 # already present — handles pre-installed AMIs without failing.
@@ -24,7 +25,6 @@ install_aws_cli() {
     /tmp/awscli-install/aws/install
     rm -rf /tmp/awscliv2.zip /tmp/awscli-install
 }
-# space-separated strings so callers can splat them with `apt_install $SET`.
 
 # ----------------------------------------------------------------------------
 # Debian family (apt)
--- a/.install/lib/packages.sh
+++ b/.install/lib/packages.sh
@@ -20,7 +20,7 @@ install_aws_cli() {
     local url="https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip"
     [ "$arch" = "aarch64" ] && url="https://awscli.amazonaws.com/awscli-exe-linux-aarch64.zip"
     curl "$url" -o /tmp/awscliv2.zip
-    unzip /tmp/awscliv2.zip -d /tmp/awscli-install
+    unzip -o /tmp/awscliv2.zip -d /tmp/awscli-install
     /tmp/awscli-install/aws/install
     rm -rf /tmp/awscliv2.zip /tmp/awscli-install
 }
--- a/.install/lib/packages.sh
+++ b/.install/lib/packages.sh
@@ -20,9 +20,9 @@ install_aws_cli() {
     arch=$(uname -m)
     local url="https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip"
     [ "$arch" = "aarch64" ] && url="https://awscli.amazonaws.com/awscli-exe-linux-aarch64.zip"
-    curl "$url" -o /tmp/awscliv2.zip
+    curl -fSL --retry 3 "$url" -o /tmp/awscliv2.zip
     unzip -o /tmp/awscliv2.zip -d /tmp/awscli-install
-    /tmp/awscli-install/aws/install
+    $SUDO /tmp/awscli-install/aws/install
     rm -rf /tmp/awscliv2.zip /tmp/awscli-install
 }
 
--- a/.install/lib/packages.sh
+++ b/.install/lib/packages.sh
@@ -19,9 +19,10 @@ install_aws_cli() {
     arch=$(uname -m)
     local url="https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip"
     [ "$arch" = "aarch64" ] && url="https://awscli.amazonaws.com/awscli-exe-linux-aarch64.zip"
-    curl "$url" -o "awscliv2.zip"
-    unzip awscliv2.zip
-    ./aws/install
+    curl "$url" -o /tmp/awscliv2.zip
+    unzip /tmp/awscliv2.zip -d /tmp/awscli-install
+    /tmp/awscli-install/aws/install
+    rm -rf /tmp/awscliv2.zip /tmp/awscli-install
 }
 # space-separated strings so callers can splat them with `apt_install $SET`.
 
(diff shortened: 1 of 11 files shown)


.install/os/bionic.sh
```diff
@@ -1,35 +1,45 @@
...
+# (ESM repos already carry gcc-10 on Ubuntu Pro machines).
 
 # shellcheck source=../lib/packages.sh
 . "$LIB/packages.sh"
 
+export DEBIAN_FRONTEND=noninteractive
+echo 'debconf debconf/frontend select Noninteractive' | $SUDO debconf-set-selections 2>/dev/null || true
+echo 'tzdata tzdata/Areas select Etc' | $SUDO debconf-set-selections 2>/dev/null || true
+echo 'tzdata tzdata/Zones/Etc select UTC' | $SUDO debconf-set-selections 2>/dev/null || true
+$SUDO apt-get install -y --no-install-recommends gnupg wget 2>/dev/null || true
```
gabsow: `gnupg`/`wget` are installed here before any `apt-get update` has run. `apt_install`'s deferred `apt-get update` (guarded by `_pm_apt_updated`) is first triggered later at `apt_install software-properties-common ...`, so on a clean bionic image — `ubuntu:18.04` ships with empty `/var/lib/apt/lists` and no preinstalled `gnupg`/`wget` — this `apt-get install` finds nothing to resolve and, because it ends in `2>/dev/null || true`, the failure is swallowed. With `gpg` then absent, the two `wget ... | gpg --dearmor ... || true` key-import lines below also silently no-op, so the toolchain-r keyrings are never written. The PPA `.list` gets added unsigned, and the later `apt_install gcc-10 g++-10` (no `|| true`) fails because apt won't install from the unsigned repo. Run `apt-get update` first and hard-install the prerequisites:

```sh
$SUDO apt-get update -qq
$SUDO apt-get install -y --no-install-recommends gnupg wget ca-certificates
```

.install/lib/packages.sh
```diff
@@ -10,6 +10,22 @@
...
+# already present — handles pre-installed AMIs without failing.
+install_aws_cli() {
+    if command -v aws &>/dev/null; then
+        return 0
+    fi
+    local arch
+    arch=$(uname -m)
+    local url="https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip"
+    [ "$arch" = "aarch64" ] && url="https://awscli.amazonaws.com/awscli-exe-linux-aarch64.zip"
+    curl "$url" -o /tmp/awscliv2.zip
```
gabsow: `curl "$url" -o /tmp/awscliv2.zip` lacks `-f`: on an HTTP 4xx/5xx, curl exits 0 and writes the error body into the file. The next line's `unzip` then fails with a confusing "cannot find zipfile directory" and, under the script's `set -euo pipefail`, aborts the whole bootstrap pointing at unzip instead of the real download failure. Use `curl -fSL --retry 3 "$url" -o /tmp/awscliv2.zip` so a bad download fails loudly at the curl step (`--retry` also rides out transient network blips).

gabsow (commented):
Reviewed the bootstrap/build changes — two robustness notes inline, both non-blocking. Heads-up on cross-module consistency: this PR's `bionic.sh` diverges from the sibling RedisJSON #1612, which runs `apt-get update` first and *hard*-installs the prerequisites (`gnupg wget curl ca-certificates`) rather than soft-failing them. Worth keeping the three modules' bootstrap in sync. (cursor's GCC-alternative points look valid too.)

gabsow:
**Cross-module consistency (non-blocking)** — this is one of three sibling PRs applying the same CI bootstrap/build fix:

- RedisJSON https://github.com/RedisJSON/RedisJSON/pull/1612
- RedisBloom https://github.com/RedisBloom/RedisBloom/pull/1028
- RedisTimeSeries https://github.com/RedisTimeSeries/RedisTimeSeries/pull/2082

Their `.install/` logic has drifted in a few places — worth reconciling so the three modules keep an identical bootstrap:

1. **bionic prerequisite install** (the one that matters): RedisJSON runs `apt-get update -qq` and then *hard*-installs `gnupg wget curl ca-certificates`; RedisBloom/RedisTimeSeries skip the update and soft-fail (`apt-get install ... gnupg wget 2>/dev/null || true`). On a clean image that can silently leave the `ubuntu-toolchain-r` PPA key unimported. Recommend standardizing on the RedisJSON variant.
2. **`debian_default_install` gcc/cc/g++ pinning** (`lib/pm.sh`): present in RedisBloom/RedisTimeSeries, absent in RedisJSON. Plausibly intentional (JSON is Rust; the others are C with `-fblocks` defer macros, so pinning gcc over clang matters there) — but worth a deliberate call rather than drift.
3. **el8 `python3.11-pip` + `update-alternatives --set python3`**: present in RedisBloom/RedisTimeSeries, absent in RedisJSON.

None of this blocks — flagging so the three don't silently diverge.

.install/os/bionic.sh
TalBarYakar: done

.install/lib/packages.sh
TalBarYakar: done

gabsow (approved):
Re-reviewed after the fix commits — all blockers resolved:

- **HIGH (GCC-10 never selected on Ubuntu):** bionic/focal now `--set` gcc/g++/**cc** to 10, so `debian_default_install`'s priority-100 pin no longer wins and the toolchain is consistently on gcc-10. ✅
- **gnupg/wget before `apt-get update`:** now runs `apt-get update` first and hard-installs the prerequisites (no more silent key-import no-op). ✅
- **curl `-f` on AWS CLI download, AL2 dangling devtoolset symlinks:** resolved. ✅

One non-blocking nit remains (not gating approval): `tests/flow/test_cuckoo.py:35` still wraps the large `CF.INSERT` in `try/except ResponseError: pass`, so the overflow test passes regardless of outcome — worth tightening in a follow-up so it can't mask a regression.

LGTM — approving.

PR #1032, opened 2026-07-10
External Tdigest Merge function

pankajkumarr-flip:
A new feature that allows user to send an external tdigest made outside redis , to be merged with an existing redis in tdigest

---

> [!NOTE]
> **High Risk**
> Touches memory sizing, untrusted binary input (EXTERNALMERGE blobs), and RDB/RESTORE loading where prior bugs could cause server crashes; behavior changes include rejecting oversized compression and mismatched RDB cap fields.
> 
> **Overview**
> Adds **`TDIGEST.EXTERNALMERGE {key} {blob}`** so client-built t-digests in the **TDB1** wire format (magic, compression, sorted centroids with weights, CRC32) can be merged into a Redis key—creating the key from the blob’s compression when missing, or keeping the existing key’s compression when it already exists. Merges run in a scratch digest and only commit on success so bad blobs never leave partial state.
> 
> Introduces shared **`TD_MAX_COMPRESSION` / `TD_MAX_CENTROIDS`** limits on **`TDIGEST.CREATE`**, **`TDIGEST.MERGE`**, and blob parsing so huge node-array allocations cannot be forced via compression alone.
> 
> **`TDigestRdbLoad`** now rejects out-of-range compression before allocation, requires the on-wire **`cap`** to match what compression actually allocates, and bounds **`merged_nodes` + `unmerged_nodes`** against that cap—closing a crafted **RESTORE**/RDB heap overflow path. Flow tests cover external merge behavior, malformed blobs, and tampered dumps.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 710128c08145abb770f3416d9f222e5236849926. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/rm_tdigest.c
+++ b/src/rm_tdigest.c
@@ -17,7 +17,9 @@
 #include "common.h"
 
 #include <math.h>
+#include <stdint.h>
 #include <stdlib.h>
+#include <string.h>
 #include <strings.h>
 #include <stdbool.h>
 
@@ -32,6 +34,14 @@
     RedisModule_TryRealloc ? RedisModule_TryRealloc(__VA_ARGS__) : RedisModule_Realloc(__VA_ARGS__)
 #define td_free_ RedisModule_Free
 #define TD_DEFAULT_COMPRESSION 100
+/* Real t-digest usage never needs more than a few thousand centroids
+ * (compression of 100-1000 is the documented sweet spot). TD_MAX_CENTROIDS
+ * bounds the per-digest node-array allocation (cap = 6*compression+10, see
+ * cap_from_compression() in the t-digest lib) to a sane size regardless of
+ * how the compression value arrives: TDIGEST.CREATE, TDIGEST.MERGE's
+ * COMPRESSION keyword, or an EXTERNALMERGE blob's declared compression. */
+#define TD_MAX_CENTROIDS (100000u) /* cap*16 bytes => <= ~1.6 MiB per digest */
+#define TD_MAX_COMPRESSION ((TD_MAX_CENTROIDS - 10) / 6)
 
 #include "tdigest.h"
 #include "load_io_error.h"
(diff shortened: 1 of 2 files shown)


LiorKogan:
@pankajkumarr-flip thank you.

I'm not sure if we need such a command. We don't have similar commands for the other probabilistic data structures, and AFAIK, we haven't seen such a requirement.

Pankaj0304:
Thanks for the feedback.

The motivation behind this command is the distributed nature of how TDigest is often used, rather than simply providing another way to update a digest.

In many production deployments, Redis runs as a centralized data store while application instances process independent streams of data. Instead of sending millions of individual "TDIGEST.ADD" commands over the network, each application instance can maintain a local TDigest, periodically serialize it, and send the compact digest to Redis for merging.

This changes the amount of work Redis performs from O(events) command processing to O(flushes) merge operations, where a single flush can represent thousands or even millions of observations. In practice, this significantly reduces network traffic, command parsing overhead, and CPU utilization on the Redis server while preserving the statistical accuracy of the aggregated TDigest.

More importantly, this aligns with one of the core design goals of the TDigest algorithm: independently constructed digests are meant to be merged. Unlike Bloom filters or several of the other probabilistic data structures, TDigest's mergeability is a fundamental property of the data structure itself. This command simply exposes that capability to support distributed aggregation workflows.

The implementation was also designed to minimize risk. Blobs are fully validated before use, merges are performed in a scratch digest and committed only on success, and strict bounds are enforced for compression, centroid counts, and RDB loading to prevent malformed or oversized inputs from affecting server stability.

I understand the concern about introducing a new public command, but I believe this enables a genuine distributed TDigest workflow that currently requires replaying every individual observation through Redis, even when those observations have already been summarized into a valid TDigest.

If the concern is primarily about the public API surface rather than the implementation itself, I'd also be happy to discuss alternative interfaces that could support the same distributed aggregation use case.

LiorKogan:
Wouldn't [DUMP](https://redis.io/docs/latest/commands/dump/) / [RESTORE](https://redis.io/docs/latest/commands/restore/) work for this use case? First restore, then merge.

If you are constructing t-digests with another library, you can extend it to support Redis' existing t-digest dump format.

PR #1037, opened 2026-07-22
MOD-16952 - enhance bootstrap (add list and dry-run arg)

TalBarYakar:
> [!NOTE]
> **Medium Risk**
> Touches all OS bootstrap paths and privileged install/repo/compiler wiring; mistakes could break fresh installs or mis-report deps, though list/dry-run are additive and normal `make bootstrap` behavior is intended to stay the same with safer re-runs.
> 
> **Overview**
> Adds **`make bootstrap list`** and **`make bootstrap dry-run`** so developers and CI can inspect bootstrap without installing packages or recreating `venv`.
> 
> **List mode** (`CHECK_DEPS=1`) probes installed packages per PM, splits **required** vs **optional** deps (`OPTIONAL_PKGS`), supports optional version floors (`MIN_VERSIONS`), and can append machine-readable lines to `DEPS_REPORT_FILE` for multi-module aggregation. It exits non-zero only when required deps are missing.
> 
> **Dry-run** prints copy-pasteable install commands for gaps only (via `_run` / `_sh`), including apt `DEBIAN_FRONTEND` and stdin redirects for pkg managers. **Real bootstrap** is unchanged except OS scripts are more **idempotent** (skip PPA/repo/toolset/symlink steps when already satisfied) and EL8 avoids false `python3` missing checks via `RHEL_PYTHON`.
> 
> Also: root containers without `sudo` get a pass-through `sudo` function; `install_aws_cli` and macOS profile/`GITHUB_PATH` updates are suppressed in list/dry-run; gcc toolset wiring won’t downgrade a newer compiler another module installed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d052e343b132ae465398863dbc5843d374a8b434. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.install/lib/packages.sh
+++ b/.install/lib/packages.sh
@@ -36,7 +36,9 @@ install_aws_cli() {
             DEPS_OPT_OK="$DEPS_OPT_OK awscli"
         else
             DEPS_OPT_MISSING="$DEPS_OPT_MISSING awscli"
-            [ "${DRY_RUN:-0}" = 1 ] && _dry_line "curl -fSL <awscli-v2>.zip && unzip && \$SUDO ./aws/install"
+            # dry-run: print the actual (arch-aware) install as one runnable line.
+            # $(uname -m) yields x86_64/aarch64, matching the AWS asset names.
+            [ "${DRY_RUN:-0}" = 1 ] && _dry_line 'curl -fSL --retry 3 "https://awscli.amazonaws.com/awscli-exe-linux-$(uname -m).zip" -o /tmp/awscliv2.zip && unzip -o /tmp/awscliv2.zip -d /tmp/awscli-install && sudo /tmp/awscli-install/aws/install && rm -rf /tmp/awscliv2.zip /tmp/awscli-install'
         fi
         return 0
     fi
--- a/.install/lib/packages.sh
+++ b/.install/lib/packages.sh
@@ -79,7 +79,6 @@ RHEL_BASE="
     openssl openssl-devel bzip2-devel libffi-devel zlib-devel
     clang clang-devel
     tcl
-    python3 python3-pip python3-devel python3-numpy
     cmake
     unzip rsync valgrind lcov jq tar which gdb
 "
--- a/.install/lib/packages.sh
+++ b/.install/lib/packages.sh
@@ -15,7 +15,7 @@
 # Only affects `make bootstrap check-deps` (reported separately, never fails).
 # Shared superset across modules; names this module doesn't install are simply
 # never matched.
-OPTIONAL_PKGS="valgrind gdb lcov tcl jq clang-format py3-numpy python3-numpy py3-psutil python3-psutil py3-cryptography python3-cryptography openssh xsimd openblas-dev curl tar uv"
+OPTIONAL_PKGS="valgrind gdb lcov tcl jq clang-format py3-numpy python3-numpy py3-psutil python3-psutil py3-cryptography python3-cryptography openssh xsimd openblas-dev curl tar uv zlib1g-dev zlib-dev zlib-devel libbz2-dev bzip2-dev bzip2-devel libblocksruntime-dev libev-dev libev-devel libevent-dev libevent-devel"
 
 # Install AWS CLI v2 from the official installer (arch-aware). Skips if
 # already present — handles pre-installed AMIs without failing.
--- a/.install/lib/packages.sh
+++ b/.install/lib/packages.sh
@@ -15,7 +15,7 @@
 # Only affects `make bootstrap list` (reported separately, never fails).
 # Shared superset across modules; names this module doesn't install are simply
 # never matched.
-OPTIONAL_PKGS="valgrind gdb lcov tcl jq clang-format py3-numpy python3-numpy py3-psutil python3-psutil py3-cryptography python3-cryptography openssh xsimd openblas-dev curl tar uv zlib1g-dev zlib-dev zlib-devel libbz2-dev bzip2-dev bzip2-devel libblocksruntime-dev libev-dev libev-devel libevent-dev libevent-devel"
+OPTIONAL_PKGS="valgrind gdb lcov tcl jq clang-format py3-numpy python3-numpy py3-psutil python3-psutil py3-cryptography python3-cryptography openssh xsimd openblas-dev curl tar uv zlib1g-dev zlib-dev zlib-devel libbz2-dev bzip2-dev bzip2-devel libblocksruntime-dev libev-dev libev-devel libevent-dev libevent-devel awscli"
 
 # MIN_VERSIONS: sparse "pkg:minversion" list — only deps with a real floor a
 # package-manager install can actually satisfy. Everything else is presence-only.
@@ -28,6 +28,18 @@ MIN_VERSIONS=""
 # Install AWS CLI v2 from the official installer (arch-aware). Skips if
 # already present — handles pre-installed AMIs without failing.
 install_aws_cli() {
+    # aws-cli is an optional CI artifact-upload tool, not a build/run dep.
+    # list/dry-run must not mutate: the curl+unzip below hit the network and
+    # write /tmp, so short-circuit before them. list records it as optional.
+    if [ "${CHECK_DEPS:-0}" = 1 ] || [ "${DRY_RUN:-0}" = 1 ]; then
+        if command -v aws >/dev/null 2>&1; then
+            DEPS_OPT_OK="$DEPS_OPT_OK awscli"
+        else
+            DEPS_OPT_MISSING="$DEPS_OPT_MISSING awscli"
+            [ "${DRY_RUN:-0}" = 1 ] && _dry_line "curl -fSL <awscli-v2>.zip && unzip && \$SUDO ./aws/install"
+        fi
+        return 0
+    fi
     if command -v aws &>/dev/null; then
         return 0
     fi
--- a/.install/lib/packages.sh
+++ b/.install/lib/packages.sh
@@ -12,7 +12,7 @@
 
 # Optional = installed by bootstrap but NOT in the README's minimal build-dep
 # list: tests/coverage/debug and feature libs the core build/run doesn't need.
-# Only affects `make bootstrap check-deps` (reported separately, never fails).
+# Only affects `make bootstrap list` (reported separately, never fails).
 # Shared superset across modules; names this module doesn't install are simply
 # never matched.
 OPTIONAL_PKGS="valgrind gdb lcov tcl jq clang-format py3-numpy python3-numpy py3-psutil python3-psutil py3-cryptography python3-cryptography openssh xsimd openblas-dev curl tar uv zlib1g-dev zlib-dev zlib-devel libbz2-dev bzip2-dev bzip2-devel libblocksruntime-dev libev-dev libev-devel libevent-dev libevent-devel"
--- a/.install/lib/packages.sh
+++ b/.install/lib/packages.sh
@@ -15,7 +15,7 @@
 # Only affects `make bootstrap check-deps` (reported separately, never fails).
 # Shared superset across modules; names this module doesn't install are simply
 # never matched.
-OPTIONAL_PKGS="valgrind gdb lcov tcl jq py3-numpy py3-psutil py3-cryptography openssh xsimd openblas-dev"
+OPTIONAL_PKGS="valgrind gdb lcov tcl jq clang-format py3-numpy python3-numpy py3-psutil python3-psutil py3-cryptography python3-cryptography openssh xsimd openblas-dev curl tar uv"
 
 # Install AWS CLI v2 from the official installer (arch-aware). Skips if
 # already present — handles pre-installed AMIs without failing.
--- a/.install/lib/packages.sh
+++ b/.install/lib/packages.sh
@@ -10,6 +10,13 @@
 # Sourced by os/<osnick>.sh after lib/pm.sh. All variables here are plain
 # space-separated strings so callers can splat them with `apt_install $SET`.
 
+# Optional = installed by bootstrap but NOT in the README's minimal build-dep
+# list: tests/coverage/debug and feature libs the core build/run doesn't need.
+# Only affects `make bootstrap check-deps` (reported separately, never fails).
+# Shared superset across modules; names this module doesn't install are simply
+# never matched.
+OPTIONAL_PKGS="valgrind gdb lcov tcl jq py3-numpy py3-psutil py3-cryptography openssh xsimd openblas-dev"
+
 # Install AWS CLI v2 from the official installer (arch-aware). Skips if
 # already present — handles pre-installed AMIs without failing.
 install_aws_cli() {
--- a/.install/lib/packages.sh
+++ b/.install/lib/packages.sh
@@ -17,6 +17,11 @@
 # never matched.
 OPTIONAL_PKGS="valgrind gdb lcov tcl jq clang-format py3-numpy python3-numpy py3-psutil python3-psutil py3-cryptography python3-cryptography openssh xsimd openblas-dev curl tar uv zlib1g-dev zlib-dev zlib-devel libbz2-dev bzip2-dev bzip2-devel libblocksruntime-dev libev-dev libev-devel libevent-dev libevent-devel"
 
+# MIN_VERSIONS: sparse "pkg:minversion" -- only deps with a real floor; a
+# present-but-older dep is reported as missing, tagged with the version.
+# cmake 3.25 = what install_cmake.sh provisions (apt 3.22 is too old).
+MIN_VERSIONS="cmake:3.25"
+
 # Install AWS CLI v2 from the official installer (arch-aware). Skips if
 # already present — handles pre-installed AMIs without failing.
 install_aws_cli() {
--- a/.install/lib/packages.sh
+++ b/.install/lib/packages.sh
@@ -17,10 +17,13 @@
 # never matched.
 OPTIONAL_PKGS="valgrind gdb lcov tcl jq clang-format py3-numpy python3-numpy py3-psutil python3-psutil py3-cryptography python3-cryptography openssh xsimd openblas-dev curl tar uv zlib1g-dev zlib-dev zlib-devel libbz2-dev bzip2-dev bzip2-devel libblocksruntime-dev libev-dev libev-devel libevent-dev libevent-devel"
 
-# MIN_VERSIONS: sparse "pkg:minversion" -- only deps with a real floor; a
-# present-but-older dep is reported as missing, tagged with the version.
-# cmake 3.25 = what install_cmake.sh provisions (apt 3.22 is too old).
-MIN_VERSIONS="cmake:3.25"
+# MIN_VERSIONS: sparse "pkg:minversion" list — only deps with a real floor a
+# package-manager install can actually satisfy. Everything else is presence-only.
+# (cmake is intentionally NOT here: its >= 3.25 requirement is provisioned by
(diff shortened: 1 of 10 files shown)


.install/lib/pm.sh
```diff
@@ -34,6 +34,108 @@ else
...
+}
+
+# DRY_RUN=1: run the bootstrap flow but install nothing — for each MISSING
+# package, print the exact install command that WOULD run (the same single
+# line the primitive executes normally, via _run — no duplicated command).
+DRY_RUN="${DRY_RUN:-0}"
+
+if [ "$CHECK_DEPS" = 1 ] || [ "$DRY_RUN" = 1 ]; then
+    _SUDO_DISPLAY="$SUDO"   # remember the real sudo prefix for dry-run printing
+    SUDO=":"                # neutralize privileged side-commands (no mutation)
```
gabsow, line 95: **`list`/`dry-run` are not actually read-only, and both under-report — the safety model misses the per-OS scripts.**

The mode machinery neutralizes only two things: `$SUDO` → `:` and the `*_install` primitives. Any command in an `os/<osnick>.sh` file (or `lib/packages.sh`) that touches the network/disk/pip **directly** bypasses both and runs for real under `make bootstrap list` / `dry-run`:

- `install_aws_cli` (`lib/packages.sh`) — `curl … -o /tmp/awscliv2.zip` + `unzip` execute; only the trailing `$SUDO …/aws/install` is neutralized. Reached from `os/rocky8.sh`, `rocky9.sh`, `mariner2.sh`, `azurelinux3.sh`, `bionic.sh`.
- `os/bionic.sh` — builds CMake from source (`wget` + `tar` + `./configure` + `make -j`) and runs `pip3 install dataclasses` (redis `scripts/bootstrap.sh` exports `PIP_BREAK_SYSTEM_PACKAGES=1`), plus two keyserver `wget`s.

Two more consequences of the same design:
- **Abort on a fresh host:** `os/focal.sh:12` / `os/bionic.sh:27` run `_cur=$(gcc -dumpversion | cut -d. -f1)` ungated. In check mode gcc was only *recorded*, not installed, so on a clean `ubuntu:20.04`/`18.04` the pipeline exits 127 and `set -euo pipefail` kills the script **before any report prints** — the exact fresh-host case `list` targets. (Same shape: `_pm_enable_el8_extras` does `return 1` on subscription-less RHEL 8.)
- **Under-reporting:** because `SUDO=":"` swallows every `$SUDO`-prefixed non-primitive, `list` never records `epel-release`, the `Development Tools` groupinstall, or the gcc-toolset symlinks/`update-alternatives`, and `dry-run`'s "commands above are what bootstrap would run" is false for those lines.

Probably needs a mode-aware wrapper (a `_run`-style guard, or gating) around the raw `curl`/`wget`/`pip`/`make`/tool-probe calls in `os/*.sh`, not just `$SUDO`.

.install/os/macos.sh
```diff
@@ -26,27 +26,32 @@ fi
 
 brew_default_install
 
+# python@3.11 only when the host has no python3 (see header). Skip under list;
+# print (don't run) under dry-run.
 if ! command -v python3 >/dev/null 2>&1; then
-    echo "==> [redisbloom] python3 not on PATH; installing brew python@3.11"
-    HOMEBREW_NO_AUTO_UPDATE=1 brew install python@3.11
+    HOMEBREW_NO_AUTO_UPDATE=1 _run brew install python@3.11
```
gabsow: **macOS `list` never reports a missing `python3` (dry-run does) — the two modes disagree.**

`python` is deliberately absent from `BREW_BASE`, and this conditional routes through `_run`, which is a no-op under `CHECK_DEPS` that records nothing into `DEPS_*`. So on a Mac with no `python3`, `make bootstrap list` reports everything present and exits 0, while `make bootstrap dry-run` *does* print `brew install python@3.11`. `list` should record python3 as a (missing) dependency so it matches dry-run and the real install. (Minor, same line: the dry-run print drops the `HOMEBREW_NO_AUTO_UPDATE=1` prefix, so the copy-paste command differs from what bootstrap runs.)

.install/os/macos.sh
```diff
@@ -26,27 +26,32 @@ fi
...
 update_profile() {
     local profile_path=$1
-    local newpath="export PATH=$COREUTILS:$LLVM:$GNUBIN:\$PATH"
-    grep -qxF "$newpath" "$profile_path" 2>/dev/null \
-        || printf '%s\n' "$newpath" >> "$profile_path"
+    grep -qxF "$NEWPATH" "$profile_path" 2>/dev/null \
+        || printf '%s\n' "$NEWPATH" >> "$profile_path"
     if [ -n "${GITHUB_PATH:-}" ]; then
-        printf '%s\n' "$newpath" >> "$GITHUB_PATH"
+        printf '%s\n' "$NEWPATH" >> "$GITHUB_PATH"
```
gabsow: **This writes the literal `export PATH=…` line into `$GITHUB_PATH` — wrong format.** GitHub Actions treats each line appended to `$GITHUB_PATH` as a *directory* to prepend, not shell to eval. So on a macOS runner this adds one garbage entry (`export PATH=…:$PATH`) and never actually puts `coreutils`/`llvm@18`/`gnubin` on PATH for later steps — they fall back to Apple BSD tools / clang-15. It also fires 0/1/2 times depending on how many of `~/.bash_profile`/`~/.zshrc` exist, since the write lives inside `update_profile`. RedisJSON's sibling PR has the correct form (writes the three dirs one-per-line, once, with an explaining comment) — worth copying it here (and into RedisTimeSeries, which has the same bug).

.install/lib/packages.sh
```diff
@@ -10,6 +10,21 @@
...
+OPTIONAL_PKGS="valgrind gdb lcov tcl jq clang-format py3-numpy python3-numpy py3-psutil python3-psutil py3-cryptography python3-cryptography openssh xsimd openblas-dev curl tar uv zlib1g-dev zlib-dev zlib-devel libbz2-dev bzip2-dev bzip2-devel libblocksruntime-dev libev-dev libev-devel libevent-dev libevent-devel"
+
+# MIN_VERSIONS: sparse "pkg:minversion" list — only deps with a real floor a
+# package-manager install can actually satisfy. Everything else is presence-only.
+# (cmake is intentionally NOT here: its >= 3.25 requirement is provisioned by
+# install_cmake.sh during a real `make bootstrap`, not by the distro package, so
+# a version floor here would be unsatisfiable by the apt/dnf `cmake` and would
+# only make `list`/`dry-run` disagree.)
+MIN_VERSIONS=""
+
```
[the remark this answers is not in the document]
TalBarYakar: fixed

.install/lib/pm.sh
TalBarYakar: Already addressed — gcc probe made non-fatal (|| echo 0) so list no longer aborts before the report; bionic's cmake-from-source, pip3, and keyserver wgets are gated behind CHECK_DEPS/DRY_RUN. The $SUDO→: under-reporting stays as documented behavior. (Same fix as the RedisTimeSeries PR.)

.install/lib/pm.sh
```diff
@@ -34,6 +34,108 @@ else
...
+            _min=$(_min_for "$_p")
+            if ! _pkg_installed "$_p"; then
+                if [ -n "$_min" ]; then DEPS_MISSING="$DEPS_MISSING $_p:$_min"; else DEPS_MISSING="$DEPS_MISSING $_p"; fi
+            elif [ -n "$_min" ] && _have=$(_get_installed_version "$_p") && [ -n "$_have" ] && ! version_ge "$_have" "$_min"; then
+                DEPS_MISSING="$DEPS_MISSING $_p:$_min"
+            else
+                DEPS_OK="$DEPS_OK $_p"
+            fi
+        fi
+    done
```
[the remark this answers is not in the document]
TalBarYakar: These names are in the *_BASE install superset but aren't actual build/link requirements — the module builds and runs without them (verified), so classifying them optional is intentional: list should only fail on deps that genuinely break the build/run. OPTIONAL_PKGS deliberately contains only non-build-critical names (test/coverage/debug + install-superset extras), so the _is_optional-first ordering is safe here.

.install/os/macos.sh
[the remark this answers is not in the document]
TalBarYakar: fixed

TalBarYakar: fixed

.install/os/macos.sh
TalBarYakar: Fixed — $GITHUB_PATH now gets the three directories one-per-line ($COREUTILS, $LLVM, $GNUBIN), written once outside update_profile, so GNU coreutils/llvm@18/gnubin actually land on PATH for later CI steps instead of a single garbage export PATH=… entry. Matched RedisJSON's form and applied the same fix to RedisTimeSeries.

.install/lib/pm.sh
gabsow: Main items here are fixed in the latest commit ✅ — the `gcc -dumpversion` probe no longer aborts (`… 2>/dev/null | … || echo 0`), and the raw `install_aws_cli` / cmake-from-source / `pip3 install` / keyserver-`wget` mutations are now gated (dry-run prints them; aws-cli is recorded as optional).

One minor, non-blocking gap remains: `list`/`dry-run` still don't record the `$SUDO`-neutralized side-commands on EL8/9 — `epel-release`, the `Development Tools` groupinstall, and the gcc-toolset symlinks — so they won't appear in the report. Fine to leave as a follow-up.

.install/os/macos.sh
gabsow: Fixed ✅ — `macos.sh` now records `python@3.11` into `DEPS_OK`/`DEPS_MISSING` under list and prints the install command under dry-run, so the two modes agree.

.install/os/macos.sh
gabsow: Fixed ✅ — now writes the three dirs one-per-line to `$GITHUB_PATH` (moved out of `update_profile`, gated), matching RedisJSON.

gabsow (dismissed):
Re-reviewed after the latest commit — all my earlier review points are addressed. LGTM 🚀

.install/lib/pm.sh
```diff
@@ -34,6 +34,141 @@ else
...
+        esac
+    else ${_SUDO_DISPLAY:-$SUDO} "$@"; fi
+}
+
+# _sh 'RAW SHELL' — a mutating step that doesn't fit _run (pipes, redirects,
+# cd/&&-chains, source builds). dry-run: print it verbatim so the output stays a
+# copy-pasteable script; list: skip; else: execute via eval. Write the string
+# with a literal `sudo` (these raw steps are Linux-only, where bootstrap ensures
+# sudo) so the printed line runs as-is.
+_sh() {
```
gabsow, line 134: Minor / non-blocking: `_sh` hardcodes a literal `sudo` in its command strings, whereas `_run` uses the MODE-aware prefix (`${_SUDO_DISPLAY:-$SUDO}`). These `_sh 'sudo …'` steps replace former `$SUDO …` calls, so on a Linux-root environment run with `INSTALL_SCRIPT_MODE=""` and no `sudo` binary they'd fail (`sudo: command not found`) where the old `$SUDO`-based code ran the command directly. Not a blocker (the redis dispatcher installs sudo when root on Linux, and MODE defaults to `sudo`) — but threading the real prefix through `_sh` like `_run` does would make it MODE-consistent.

.install/lib/pm.sh
```diff
@@ -149,24 +301,30 @@ _pm_enable_el8_extras() {
...
+            # before the diagnostic below can run.
+            local rid=$(dnf repolist --all 2>/dev/null \
                 | grep -i 'codeready-builder-for-rhel-8' \
                 | grep -vi source \
                 | head -1 \
                 | awk '{print $1}')
             if [ -n "$rid" ]; then
-                $SUDO dnf config-manager --set-enabled "$rid"
+                dnf repolist --enabled 2>/dev/null | grep -q "$rid" \
+                    || _run dnf config-manager --set-enabled "$rid"
```
gabsow, line 326: Minor / non-blocking (niche): a few lines below, the `return 1` when no CodeReady repo is found makes `_pm_enable_el8_extras` exit non-zero, and `el8_default_install` calls it unguarded — so under `set -euo pipefail` that aborts `make bootstrap list` before the dependency report prints. Only reachable on a *subscription-less real RHEL 8* (Alma/Rocky report `ID=almalinux`/`rocky` and skip this branch, so CI is unaffected). A `|| true` at the call site would let `list` degrade gracefully.

gabsow (dismissed):
Re-reviewed the latest revision. The read-only routing via `_sh`/`_run`, the idempotency guards, and the arm64 Amazon Linux 2 fix all look good, and the earlier follow-up is resolved — `list`/`dry-run` now report (and print) the side-commands that were previously swallowed by `SUDO=":"`, with nothing mutating. LGTM 🚀

Left two minor, non-blocking notes inline (`_sh` hardcoded `sudo` vs MODE-aware prefix; `_pm_enable_el8_extras` `return 1` on subscription-less RHEL 8) — neither needs to hold the merge.

.install/lib/pm.sh
```diff
@@ -87,14 +198,18 @@ tdnf_install() {
...
+    _run apk add --no-cache "$@"
 }
 
 # brew exits non-zero when a formula is already installed/linked. We tolerate
 # that on rerun; real failures get caught later by the caller's feature
 # checks (compiler probes, `command -v`, ...).
 brew_install() {
     [ "$#" -gt 0 ] || return 0
+    if [ "$CHECK_DEPS" = 1 ]; then _check_pkgs "$@"; return 0; fi
+    if [ "$DRY_RUN" = 1 ]; then set -- $(_missing_only "$@"); [ "$#" -gt 0 ] || return 0; _run brew install "$@"; return 0; fi
```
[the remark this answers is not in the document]
TalBarYakar: fixed

.install/os/macos.sh
```diff
@@ -26,27 +26,43 @@ fi
...
+if command -v python3 >/dev/null 2>&1; then
+    # present — record for the list count (macOS python isn't in BREW_BASE)
+    if [ "${CHECK_DEPS:-0}" = 1 ]; then DEPS_OK="$DEPS_OK python@3.11"; fi
+elif [ "${CHECK_DEPS:-0}" = 1 ]; then
+    # list: record as missing so it matches dry-run and a real install
+    DEPS_MISSING="$DEPS_MISSING python@3.11"
+elif [ "${DRY_RUN:-0}" = 1 ]; then
+    _dry_line "HOMEBREW_NO_AUTO_UPDATE=1 brew install python@3.11"
+else
+    HOMEBREW_NO_AUTO_UPDATE=1 _run brew install python@3.11
```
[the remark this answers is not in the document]
TalBarYakar: fixed

.install/lib/pm.sh
TalBarYakar: Agreed it's minor/non-blocking. The literal sudo is deliberate (noted in the comment): these _sh steps are Linux-only and the dispatcher installs sudo when running as root, with MODE defaulting to sudo, so the root+empty-MODE+no-sudo combination doesn't occur in the bootstrap path. Threading the prefix would also drop sudo from the copy-pasteable dry-run output, so leaving as-is.

.install/lib/pm.sh
TalBarYakar: Acknowledged — very minor. Only reachable on a subscription-less real RHEL 8 (Alma/Rocky skip the branch, so CI is unaffected), and on such a box CodeReady Builder is genuinely required for gcc-toolset-11, so the hard-fail with the "fix RHUI/subscription repos" diagnostic is acceptable. Leaving as-is.

.install/lib/pm.sh
```diff
@@ -34,6 +34,141 @@ else
...
+# with a real floor. Everything else is presence-only. Default empty.
+MIN_VERSIONS="${MIN_VERSIONS:-}"
+_min_for() { for _e in $MIN_VERSIONS; do case "$_e" in "$1:"*) echo "${_e#*:}"; return ;; esac; done; }
+
+# Read a tool's installed version from the tool itself (not the package DB --
+# e.g. cmake is overlaid into /usr/local by install_cmake.sh, so dpkg would
+# report the stale apt version).
+_get_installed_version() {
+    case "$1" in
+        gcc|g++) "$1" -dumpversion 2>/dev/null ;;
```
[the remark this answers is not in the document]
TalBarYakar: Acknowledged — low severity, leaving as-is. It's only reachable through the list path's version check, and in these modules MIN_VERSIONS="" so the gcc|g++ branch's return value is never version-compared. It would only bite on a broken/unrunnable gcc that the package DB still reports installed — a degenerate case where make bootstrap list exiting non-zero is acceptable. Not worth the change for that.

.install/lib/pm.sh
```diff
@@ -178,23 +336,32 @@ _pm_enable_el8_extras() {
 # download its own 3.12, then psutil's wheel-less aarch64 source build would
 # fail looking for Python.h that matches the wrong interpreter.
 el8_default_install() {
-    $SUDO dnf -y install epel-release
-    # Hard-fail: without codeready-builder, libatomic-devel (and other -devel
-    # pkgs) won't resolve in the dnf_install below, leaving a half-installed
-    # toolchain and a confusing "no package" error several steps later.
-    _pm_enable_el8_extras || exit 1
+    rpm -q epel-release >/dev/null 2>&1 || _run dnf -y install epel-release
+    _pm_enable_el8_extras
```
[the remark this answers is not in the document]
TalBarYakar: Same as the related thread above — acknowledged, very minor. Only reachable on subscription-less real RHEL 8 (Alma/Rocky skip the ID=rhel branch, so CI is unaffected), where CodeReady Builder is genuinely required for gcc-toolset-11. The hard abort with the CRB diagnostic is acceptable there. Leaving as-is.

.install/lib/pm.sh
gabsow, line 134: (Re-posting after the rebase orphaned it — non-blocking.) `_sh` hardcodes a literal `sudo`, whereas `_run` uses the MODE-aware prefix (`${_SUDO_DISPLAY:-$SUDO}`). On a Linux-root env run with `INSTALL_SCRIPT_MODE=""` and no `sudo` binary, these `_sh 'sudo …'` steps would fail where the old `$SUDO`-based code ran directly. Not a blocker (bootstrap ensures sudo when root; MODE defaults to `sudo`) — threading the real prefix through `_sh` would make it MODE-consistent.

.install/lib/pm.sh
gabsow, line 326: (Re-posting after the rebase orphaned it — non-blocking, niche.) The `return 1` when no CodeReady repo is found makes `_pm_enable_el8_extras` exit non-zero, and `el8_default_install` calls it unguarded, so under `set -euo pipefail` that aborts `make bootstrap list` before the report. Only on subscription-less real RHEL 8 (Alma/Rocky skip this branch). A `|| true` at the call site would let `list` degrade gracefully.

gabsow (dismissed):
Re-approving after the rebase — I diffed every file each PR touches at my previously-approved commit vs this rebased head and they are byte-identical, so this is a pure rebase with no content change (which is what dismissed the approval). Nothing new to review. LGTM 🚀

The two minor, non-blocking notes from before still apply (re-posted inline): `_sh` literal `sudo` vs MODE-aware prefix; `_pm_enable_el8_extras` `return 1` on subscription-less RHEL 8.

gabsow (approved):

PR #1051, opened 2026-08-16
2.6 ci: fix nightly sanitizer and valgrind jobs

AvivDavid23:
The `linux-sanitizer` and `linux-valgrind` nightly jobs have both been failing on
this branch for a while ([sanitizer](https://github.com/RedisBloom/RedisBloom/actions/runs/30031487748/job/89288909563),
[valgrind](https://github.com/RedisBloom/RedisBloom/actions/runs/30031487748/job/89288920103)).
Neither failure was a module bug — `tests/flow/tests.sh` had drifted from 8.4/master,
which already solved both. `flow-linux.yml` and `.install/install_redis.sh` are
byte-identical to 8.4, so only `tests.sh` needed to change.

## Sanitizer

Aborted with `Cannot find redis-server-asan-6.2`. `setup_clang_sanitizer` looked for
`redis-server-asan-$SAN_REDIS_VER`, but `getredis --suffix asan` installs the binary as
plain `redis-server-asan`, so the name never matched and `setup_redis_server` bailed out.

The Docker image already builds redis with `SANITIZER=address`, so rebuilding redis here
was both redundant and broken. Dropped, as on 8.4 — `redis-server` from `PATH` is already
instrumented. This also drops `LSAN_OPTIONS`, which pointed at `tests/memcheck/asan.supp`,
a file that does not exist on this branch, and adds `allocator_may_return_null=1` so the
oversized-allocation hardening tests do not abort the run.

## Valgrind

Every server died at startup with:

```
==2079== FATAL: can't open suppressions file "/workspace/tests/memcheck/valgrind.supp"
```

The file is named `tests/memcheck/redis_valgrind.sup`. Because no server started, every
test class errored in `__init__`. Also dropped the `getredis --valgrind` build for the same
reason as above.

Separately, the valgrind leak summary was guarded by `if [[ $NOP != 1 && -n $SAN ]]`, so it
never ran for valgrind-only builds even though the inner condition tested for `$VG`.

## Verification

Ran in the `jammy` container locally. Sanitizer suite green across all four topologies:

| topology | result |
|---|---|
| general | 141 run / 0 failed |
| `--use-slaves` | 141 run / 0 failed |
| `--use-aof` | 141 run / 0 failed |
| `--env oss-cluster` | 141 run / 0 failed |

The local valgrind run is **not** a clean pass — see the note below.

## Open item for CI

Local valgrind runs (on **arm64**, whereas CI is x64) hit `TDIGEST.REVRANK` off-by-one
assertion failures in `test_tdigest_revrank` and `test_tdigest_rank_and_revrank`, e.g.
`[15, 14, 13, 10, 7, 2, -1]` vs `[15, 15, 13, 11, 7, 3, -1]`. These are numeric, not memory
errors. I could not attribute them confidently: `VG=1` also implies `DEBUG=1` and adds
`-D_VALGRIND`, and valgrind's FP emulation differs by architecture. The x64 valgrind job on
this PR should settle whether it is real or an arm64 artifact.

## Review follow-ups

- `sbin/memcheck-summary` matched five specific ASan error names, so `heap-buffer-overflow`,
  `global-buffer-overflow`, `stack-buffer-overflow`, `use-after-poison` and
  `alloc-dealloc-mismatch` went unreported — and with `-fsanitize-recover=all` on the module
  plus `halt_on_error=0`, this summary is the only thing that can fail the job on those. Now
  matches the report header, `": [A-Za-z]*Sanitizer: "`. The trailing `": "` keeps out
  `WARNING: AddressSanitizer failed to allocate`, which the oversized-allocation hardening
  tests provoke on purpose. Checked against 609 real `*.asan.log*` files from local runs.
- `tests.sh` read `$CLUSTER`/`$QUICK`, but the Makefile exports `OSS_CLUSTER`, so
  `make flow-tests OSS_CLUSTER=0` still ran the oss-cluster variant. Both branches now fall
  back to `OSS_CLUSTER`. No change for the nightly, which leaves it at the default of 1.
- The `_halfRoundUp` attempt at the `TDIGEST.REVRANK` divergence has been **reverted** — over
  the reachable domain it is bit-for-bit `round()`, so it could not have changed the result,
  and its `int_part >= 0.0` guard was wrong for `(-1, 0)` where `modf` yields `-0.0`. `src/` is
  now byte-identical to `2.6`; the REVRANK item below stays open and belongs in its own PR.
- The gcc-vs-clang ASan concern raised in review does not apply: `master` builds redis with the
  image's gcc under `SANITIZER=address` and pairs it with the same clang-built module, and its
  `linux-sanitizer / x64-jammy` job is green (run 31906795267 and the three nightlies before
  it). No `CC=clang` needed here.

Pre-existing, not touched here: `master`'s own `sbin/memcheck-summary` has a broken
invocation — `grep -q -e "runtime error" -e "Sanitizer" -v "failed to allocate" "$file"` parses
the pattern as a *filename*, visible as 16 × `grep: failed to allocate: No such file or
directory` in the nightly log. Needs its own ticket.

---

> [!NOTE]
> **Low Risk**
> Changes are limited to test harness scripts and log aggregation; no module runtime or security-sensitive production paths.
> 
> **Overview**
> **Aligns flow memcheck with 8.4/master** so nightly `linux-sanitizer` and `linux-valgrind` jobs can start Redis and fail correctly on log findings.
> 
> **Sanitizer:** `tests/flow/tests.sh` no longer builds or looks for `redis-server-asan-*` / MSAN variants via `getredis`; it uses `redis-server` from `PATH` (already instrumented in CI). `setup_sanitizer` sets `ASAN_OPTIONS` with `halt_on_error=0` and `allocator_may_return_null=1`, and drops `LSAN_OPTIONS` pointing at a missing suppressions file.
> 
> **Valgrind:** Drops the redundant `getredis --valgrind` path and points suppressions at `tests/memcheck/redis_valgrind.sup` instead of the non-existent `valgrind.supp`.
> 
> **Post-run summary:** `sbin/memcheck-summary` scans `tests/flow/logs` (not `pytests`) and uses one grep for `: *Sanitizer: ` so heap/global/stack buffer overflows and similar reports are not missed when the run continues past errors. Flow tests now invoke that summary for Valgrind-only runs as well as sanitizer runs.
> 
> **Topology:** `CLUSTER` defaults honor `OSS_CLUSTER` from the Makefile when `CLUSTER` is unset.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e2634ae5374f70785162171745cd78ff14c9ae47. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/rm_tdigest.c
+++ b/src/rm_tdigest.c
@@ -76,19 +76,6 @@ double _halfRoundDown(double f) {
     return int_part >= 0.0 ? int_part + 1.0 : int_part - 1.0;
 }
 
-// Counterpart of _halfRoundDown. Not expressed with round(), whose tie behaviour is
-// libc/platform dependent -- under Valgrind on aarch64 an exact .5 rounds to even
-// instead of away from zero, which silently shifts TDIGEST.REVRANK results.
-double _halfRoundUp(double f) {
-    double int_part;
-    double frac_part = modf(f, &int_part);
-
-    if (fabs(frac_part) < 0.5)
-        return int_part;
-
-    return int_part >= 0.0 ? int_part + 1.0 : int_part - 1.0;
-}
-
 /**
  * Command: TDIGEST.CREATE {key} [{compression}]
  *
@@ -492,7 +479,7 @@ static int _TDigest_Rank(RedisModuleCtx *ctx, RedisModuleString *const *argv, in
             const double cdf_val = td_cdf(tdigest, vals[i]);
             const double cdf_val_prior_round = cdf_val * size;
             const double cdf_to_absolute =
-                reverse ? _halfRoundUp(cdf_val_prior_round) : _halfRoundDown(cdf_val_prior_round);
+                reverse ? round(cdf_val_prior_round) : _halfRoundDown(cdf_val_prior_round);
             ranks[i] = reverse ? round(size - cdf_to_absolute) : cdf_to_absolute;
         }
     }
--- a/src/rm_tdigest.c
+++ b/src/rm_tdigest.c
@@ -76,6 +76,19 @@ double _halfRoundDown(double f) {
     return int_part >= 0.0 ? int_part + 1.0 : int_part - 1.0;
 }
 
+// Counterpart of _halfRoundDown. Not expressed with round(), whose tie behaviour is
+// libc/platform dependent -- under Valgrind on aarch64 an exact .5 rounds to even
+// instead of away from zero, which silently shifts TDIGEST.REVRANK results.
+double _halfRoundUp(double f) {
+    double int_part;
+    double frac_part = modf(f, &int_part);
+
+    if (fabs(frac_part) < 0.5)
+        return int_part;
+
+    return int_part >= 0.0 ? int_part + 1.0 : int_part - 1.0;
+}
+
 /**
  * Command: TDIGEST.CREATE {key} [{compression}]
  *
@@ -479,7 +492,7 @@ static int _TDigest_Rank(RedisModuleCtx *ctx, RedisModuleString *const *argv, in
             const double cdf_val = td_cdf(tdigest, vals[i]);
             const double cdf_val_prior_round = cdf_val * size;
             const double cdf_to_absolute =
-                reverse ? round(cdf_val_prior_round) : _halfRoundDown(cdf_val_prior_round);
+                reverse ? _halfRoundUp(cdf_val_prior_round) : _halfRoundDown(cdf_val_prior_round);
             ranks[i] = reverse ? round(size - cdf_to_absolute) : cdf_to_absolute;
         }
     }
--- a/tests/flow/tests.sh
+++ b/tests/flow/tests.sh
@@ -8,9 +8,6 @@ READIES=$ROOT/deps/readies
 
 export PYTHONUNBUFFERED=1
 
-VALGRIND_REDIS_VER=6.2
-SAN_REDIS_VER=6.2
-
 cd $HERE
 
 #----------------------------------------------------------------------------------------------
@@ -159,42 +156,14 @@ setup_rltest() {
 
 #----------------------------------------------------------------------------------------------
 
-setup_clang_sanitizer() {
-	local ignorelist=$ROOT/tests/memcheck/redis.san-ignorelist
-	if ! grep THPIsEnabled $ignorelist &> /dev/null; then
-		echo "fun:THPIsEnabled" >> $ignorelist
-	fi
-
-	# for RediSearch module
-	export RS_GLOBAL_DTORS=1
-
+setup_sanitizer() {
 	# for RLTest
 	export SANITIZER="$SAN"
 	export SHORT_READ_BYTES_DELTA=512
-	
+
+	export ASAN_OPTIONS="detect_odr_violation=0:halt_on_error=0:detect_leaks=1:allocator_may_return_null=1"
 	# --no-output-catch --exit-on-failure --check-exitcode
 	RLTEST_SAN_ARGS="--sanitizer $SAN"
-
-	if [[ $SAN == addr || $SAN == address ]]; then
-		REDIS_SERVER=${REDIS_SERVER:-redis-server-asan-$SAN_REDIS_VER}
-		if ! command -v $REDIS_SERVER > /dev/null; then
-			echo Building Redis for clang-asan ...
-			$READIES/bin/getredis --force -v $SAN_REDIS_VER --own-openssl --no-run \
-				--suffix asan --clang-asan --clang-san-blacklist $ignorelist
-		fi
-
-		export ASAN_OPTIONS="detect_odr_violation=0:halt_on_error=0:detect_leaks=1"
-		export LSAN_OPTIONS="suppressions=$ROOT/tests/memcheck/asan.supp"
-
-	elif [[ $SAN == mem || $SAN == memory ]]; then
-		REDIS_SERVER=${REDIS_SERVER:-redis-server-msan-$SAN_REDIS_VER}
-		if ! command -v $REDIS_SERVER > /dev/null; then
-			echo Building Redis for clang-msan ...
-			$READIES/bin/getredis --force -v $SAN_REDIS_VER  --no-run --own-openssl \
-				--suffix msan --clang-msan --llvm-dir /opt/llvm-project/build-msan \
-				--clang-san-blacklist $ignorelist
-		fi
-	fi
 }
 
 #----------------------------------------------------------------------------------------------
@@ -211,12 +180,6 @@ setup_redis_server() {
 #----------------------------------------------------------------------------------------------
 
 setup_valgrind() {
-	REDIS_SERVER=${REDIS_SERVER:-redis-server-vg}
-	if ! is_command $REDIS_SERVER; then
-		echo Building Redis for Valgrind ...
-		$READIES/bin/getredis -v $VALGRIND_REDIS_VER --valgrind --suffix vg
-	fi
-
 	if [[ $VG_LEAKS == 0 ]]; then
 		VG_LEAK_CHECK=no
 		RLTEST_VG_NOLEAKS="--vg-no-leakcheck"
@@ -232,7 +195,7 @@ setup_valgrind() {
 		--track-origins=yes \
 		--show-possibly-lost=no"
 
-	VALGRIND_SUPRESSIONS=$ROOT/tests/memcheck/valgrind.supp
+	VALGRIND_SUPRESSIONS=$ROOT/tests/memcheck/redis_valgrind.sup
 
 	RLTEST_VG_ARGS+="\
 		--use-valgrind \
@@ -526,7 +489,7 @@ fi
 setup_rltest
 
 if [[ -n $SAN ]]; then
-	setup_clang_sanitizer
+	setup_sanitizer
 fi
 
 if [[ $VG == 1 ]]; then
@@ -565,7 +528,7 @@ if [[ $NO_SUMMARY == 1 ]]; then
 	exit 0
 fi
 
-if [[ $NOP != 1 && -n $SAN ]]; then
+if [[ $NOP != 1 ]]; then
 	if [[ -n $SAN || $VG == 1 ]]; then
 		{ FLOW=1 $ROOT/sbin/memcheck-summary; (( E |= $? )); } || true
 	fi
--- a/tests/flow/tests.sh
+++ b/tests/flow/tests.sh
@@ -161,6 +161,8 @@ setup_sanitizer() {
 	export SANITIZER="$SAN"
 	export SHORT_READ_BYTES_DELTA=512
 
+	# halt_on_error=0 keeps the run going past the first report so every finding lands in
+	# logs/*.asan.log; sbin/memcheck-summary then scans them and fails the job.
 	export ASAN_OPTIONS="detect_odr_violation=0:halt_on_error=0:detect_leaks=1:allocator_may_return_null=1"
 	# --no-output-catch --exit-on-failure --check-exitcode
 	RLTEST_SAN_ARGS="--sanitizer $SAN"
@@ -364,16 +366,17 @@ OS=$($READIES/bin/platform --os)
 
 #---------------------------------------------------------------------------------- Tests scope
 
+# the Makefile exports OSS_CLUSTER, not CLUSTER, so accept both names
 if [[ $QUICK == 1 ]]; then
 	GEN=${GEN:-1}
 	SLAVES=${SLAVES:-0}
 	AOF=${AOF:-0}
-	CLUSTER=${CLUSTER:-0}
+	CLUSTER=${CLUSTER:-${OSS_CLUSTER:-0}}
 else
 	GEN=${GEN:-1}
 	SLAVES=${SLAVES:-1}
 	AOF=${AOF:-1}
-	CLUSTER=${CLUSTER:-1}
+	CLUSTER=${CLUSTER:-${OSS_CLUSTER:-1}}
 fi
 
 RLEC=${RLEC:-0}
--- a/sbin/memcheck-summary
+++ b/sbin/memcheck-summary
@@ -56,13 +56,14 @@ sanitizer_check() {
 
 sanitizer_summary() {
 	local logdir="$ROOT/tests/$DIR/logs"
-	if ! TYPE="leaks" sanitizer_check "Direct leak"; then
-		TYPE="leaks" sanitizer_check "detected memory leaks"
-	fi
-	TYPE="buffer overflow" sanitizer_check "dynamic-stack-buffer-overflow"
-	TYPE="memory errors" sanitizer_check "memcpy-param-overlap"
-	TYPE="stack use after scope" sanitizer_check "stack-use-after-scope"
-	TYPE="use after free" sanitizer_check "heap-use-after-free"
+	# Every sanitizer report is headed by "<ERROR|WARNING>: <Tool>Sanitizer: <what>", so one
+	# pattern covers all of them. The previous explicit list missed heap-buffer-overflow,
+	# global-buffer-overflow, stack-buffer-overflow, use-after-poison and alloc-dealloc-mismatch
+	# -- and since the module is built with -fsanitize-recover=all and tests.sh sets
+	# halt_on_error=0, this summary is the only thing that fails the job on those.
+	# The trailing ": " matters: it keeps out "WARNING: AddressSanitizer failed to allocate",
+	# which the oversized-allocation hardening tests provoke on purpose.
+	TYPE="errors" sanitizer_check ": [A-Za-z]*Sanitizer: "
 }
 
 #----------------------------------------------------------------------------------------------
--- a/sbin/memcheck-summary
+++ b/sbin/memcheck-summary
@@ -74,7 +74,7 @@ DIRS=
 #	DIRS+=" ctests cpptests"
 #fi
 if [[ $FLOW == 1 ]]; then
-	DIRS+=" pytests"
+	DIRS+=" flow"
 fi
 
 if [[ $VG == 1 ]]; then


src/rm_tdigest.c
```diff
@@ -479,7 +492,7 @@ static int _TDigest_Rank(RedisModuleCtx *ctx, RedisModuleString *const *argv, in
             const double cdf_val = td_cdf(tdigest, vals[i]);
             const double cdf_val_prior_round = cdf_val * size;
             const double cdf_to_absolute =
-                reverse ? round(cdf_val_prior_round) : _halfRoundDown(cdf_val_prior_round);
+                reverse ? _halfRoundUp(cdf_val_prior_round) : _halfRoundDown(cdf_val_prior_round);
```
gabsow: For the only inputs this call site can produce, `_halfRoundUp` is bit-for-bit identical to `round()`, so this change is a semantic no-op and cannot fix the reported `TDIGEST.REVRANK` off-by-ones.

`cdf_val` is in `[0,1]` and `size >= 1` (the `size == 0` case returns early), so `cdf_val_prior_round >= 0` always. For `x >= 0`, `modf` gives `int_part = trunc(x)` and `frac` in `[0,1)`:
- `frac < 0.5` -> `trunc(x)`, which is `round(x)`
- `frac >= 0.5` -> `trunc(x) + 1`, which is `round(x)`

I compiled the helper and swept `size` 1..500 x `cdf` in 4001 steps (2M points over the reachable domain): **0 mismatches vs `round()`**.

So the arm64 Valgrind failure quoted in the PR description — `[15, 14, 13, 10, 7, 2, -1]` vs `[15, 15, 13, 11, 7, 3, -1]` — will reproduce identically after this patch. Four positions differ by 1, which cannot all be exact `.5` ties, and even a genuine tie is resolved the same way by both functions.

The comment above also states an incorrect premise: C99 `round()` is specified to round halfway cases away from zero *regardless* of the current FP rounding mode, and Valgrind emulates FP arithmetic but does not substitute its own libm `round()`. Whatever is shifting REVRANK on aarch64 is upstream of this line (most likely `td_cdf`, or the `-D_VALGRIND` / `DEBUG=1` that `VG=1` implies).

Since the open item in the PR description is still unresolved, I would drop this hunk from the CI fix and keep `src/` untouched on a release branch, rather than land a change that has no effect and whose comment records a wrong diagnosis.

src/rm_tdigest.c
```diff
@@ -76,6 +76,19 @@ double _halfRoundDown(double f) {
...
+// libc/platform dependent -- under Valgrind on aarch64 an exact .5 rounds to even
+// instead of away from zero, which silently shifts TDIGEST.REVRANK results.
+double _halfRoundUp(double f) {
+    double int_part;
+    double frac_part = modf(f, &int_part);
+
+    if (fabs(frac_part) < 0.5)
+        return int_part;
+
+    return int_part >= 0.0 ? int_part + 1.0 : int_part - 1.0;
```
gabsow: Guarding on `int_part` rather than `f` makes this wrong for every negative input in `(-1, 0)`: `modf` returns `int_part = -0.0` there, and `-0.0 >= 0.0` is **true** in IEEE 754, so the function adds 1 instead of subtracting.

Verified by compiling the helper:

```
_halfRoundUp(-0.5) = 1    (round gives -1)
_halfRoundUp(-0.9) = 1    (round gives -1)
```

Not reachable from the current call site (`cdf_val * size` is never negative), but the function is non-static and its comment presents it as a general round-half-away-from-zero utility, so the next caller that can pass a negative gets the wrong sign — and the magnitude is wrong too, not just the sign. `_halfRoundDown` above has the identical defect.

```suggestion
    return f >= 0.0 ? int_part + 1.0 : int_part - 1.0;
```

tests/flow/tests.sh
```diff
@@ -159,42 +156,14 @@ setup_rltest() {
...
-	# for RediSearch module
-	export RS_GLOBAL_DTORS=1
-
+setup_sanitizer() {
 	# for RLTest
 	export SANITIZER="$SAN"
 	export SHORT_READ_BYTES_DELTA=512
-	
+
+	export ASAN_OPTIONS="detect_odr_violation=0:halt_on_error=0:detect_leaks=1:allocator_may_return_null=1"
```
gabsow, line 166: `halt_on_error=0` is new here, and combined with what `memcheck-summary` actually greps for it can let a real ASan finding through as a green nightly.

`sanitizer_summary` in `sbin/memcheck-summary` only greps for:

- `Direct leak` / `detected memory leaks`
- `dynamic-stack-buffer-overflow`
- `memcpy-param-overlap`
- `stack-use-after-scope`
- `heap-use-after-free`

Not covered: `heap-buffer-overflow`, `global-buffer-overflow`, `stack-buffer-overflow`, `use-after-poison`, `alloc-dealloc-mismatch` — `heap-buffer-overflow` in particular being the most common ASan report there is.

readies compiles the module with `-fsanitize-recover=all` (`deps/readies/mk/clang-sanitizer.defs`), so module-side findings are recoverable. Concretely: a `heap-buffer-overflow` in module code gets written to `tests/flow/logs/*.asan.log`, the server keeps running because `halt_on_error=0`, no grep in `memcheck-summary` matches, `E` stays 0, and the job reports success. Without `halt_on_error=0` ASan would have aborted the server on the first report and the test would have failed.

`--check-exitcode` is not passed to RLTest either (it is in the commented-out line just above), so ASan's exit code 23 is not a backstop.

If `halt_on_error=0` is only there for the oversized-allocation hardening tests, `allocator_may_return_null=1` already covers those — those return NULL rather than reporting an error. Worth either dropping `halt_on_error=0` or widening the grep list in `memcheck-summary` in the same PR.

tests/flow/tests.sh
```diff
@@ -159,42 +156,14 @@ setup_rltest() {
...
-setup_clang_sanitizer() {
-	local ignorelist=$ROOT/tests/memcheck/redis.san-ignorelist
-	if ! grep THPIsEnabled $ignorelist &> /dev/null; then
-		echo "fun:THPIsEnabled" >> $ignorelist
-	fi
-
-	# for RediSearch module
-	export RS_GLOBAL_DTORS=1
-
+setup_sanitizer() {
```
gabsow, line 159: Dropping the `getredis --clang-asan` build means the sanitizer job now runs the image's redis, and on 2.6 that redis is **gcc**-built while the module is **clang**-built — which is exactly the mismatch the companion PR #1052 diagnoses as fatal.

- `deps/readies/mk/clang-sanitizer.defs` forces `CLANG=1` whenever `SAN` is set, so `redisbloom.so` is clang-built with `-fsanitize=address` and links no ASan runtime of its own.
- 2.6's `Dockerfile.jammy` sets no `CC`, and 2.6's `.install/install_redis.sh` is unchanged (`make SANITIZER=${SANITIZER:-} -j$(nproc)`), so redis 7.2 is built with gcc and the process ASan runtime is gcc's `libasan.so.6`.
- The module then resolves its `__asan_*` symbols against gcc's runtime.

That is the configuration #1052 reports as aborting the server at module load with a bogus `global-buffer-overflow` in `RM_GetApi`, and #1052 fixes it by passing `CC=clang LD=clang` to redis's make. This PR does not.

The two companion PRs are inconsistent on this point, and only one of the two readings can be right:
- if the mismatch is real, 2.6 needs the same `CC=clang LD=clang` change (this PR's green local run was on **arm64**; the nightly is x64, and clang/gcc ASan ABI compatibility is exactly the kind of thing that differs there);
- if it is not real, the `CC=clang` change in #1052 is unnecessary and its comment is misleading.

Worth settling before either lands, since the answer changes what the other PR should look like.

src/rm_tdigest.c
AvivDavid23: Agreed, and dropped — `src/` is now byte-identical to the base branch on both PRs (revert commits e2634ae on #1051, cf19ca5 on #1052). Confirmed with `git diff origin/<base>...HEAD -- src/` returning empty.

Your analysis holds on both points: over the reachable domain (`cdf_val` in `[0,1]`, `size >= 1`) `_halfRoundUp` is `round()`, and the `int_part >= 0.0` guard is wrong for `(-1, 0)` because `modf` yields `-0.0` there. So the patch could not have moved the arm64 REVRANK result, and the comment's premise about C99 `round()` was wrong.

The arm64 Valgrind REVRANK divergence stays open and moves out of the CI fix — it needs to be chased upstream of this line (`td_cdf`, or the `-D_VALGRIND` / `DEBUG=1` that `VG=1` implies), not in the rounding helper.

src/rm_tdigest.c
AvivDavid23: Agreed, and dropped — `src/` is now byte-identical to the base branch on both PRs (revert commits e2634ae on #1051, cf19ca5 on #1052). Confirmed with `git diff origin/<base>...HEAD -- src/` returning empty.

Your analysis holds on both points: over the reachable domain (`cdf_val` in `[0,1]`, `size >= 1`) `_halfRoundUp` is `round()`, and the `int_part >= 0.0` guard is wrong for `(-1, 0)` because `modf` yields `-0.0` there. So the patch could not have moved the arm64 REVRANK result, and the comment's premise about C99 `round()` was wrong.

The arm64 Valgrind REVRANK divergence stays open and moves out of the CI fix — it needs to be chased upstream of this line (`td_cdf`, or the `-D_VALGRIND` / `DEBUG=1` that `VG=1` implies), not in the rounding helper.

tests/flow/tests.sh
```diff
@@ -565,7 +528,7 @@ if [[ $NO_SUMMARY == 1 ]]; then
 	exit 0
 fi
 
-if [[ $NOP != 1 && -n $SAN ]]; then
+if [[ $NOP != 1 ]]; then
```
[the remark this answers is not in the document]
AvivDavid23: Already fixed in f18beb2, which landed before this review: `DIRS+=" flow"` in `sbin/memcheck-summary`, so the valgrind summary scans `tests/flow/logs` where RLTest actually writes.

Worth recording that the dispatch does work: `tests.sh` invokes the summary as `FLOW=1 $ROOT/sbin/memcheck-summary`, and `$VG` / `$SAN` are not `export`ed anywhere in `tests.sh` — but `flow-linux.yml` runs `make test VG=... SAN=...`, and GNU make puts command-line variables into the environment of every recipe, so both reach the child. With the `-n $SAN` guard widened and `DIRS` corrected, `--vg-no-fail-on-errors` is now genuinely backed by a post-run gate.

tests/flow/tests.sh
AvivDavid23: Confirmed and fixed, by widening the grep rather than dropping `halt_on_error=0` (29cad35).

`sanitizer_summary` now matches the report header instead of an enumeration:

```bash
TYPE="errors" sanitizer_check ": [A-Za-z]*Sanitizer: "
```

Every sanitizer report is headed by `<ERROR|WARNING>: <Tool>Sanitizer: <what>`, so this covers `heap-buffer-overflow`, `global-buffer-overflow`, `stack-buffer-overflow`, `use-after-poison` and `alloc-dealloc-mismatch` as well as the five that were listed, and it keeps working for MSan/TSan.

The trailing `": "` is load-bearing: with `allocator_may_return_null=1`, an oversized allocation prints `==N==WARNING: AddressSanitizer failed to allocate 0x... bytes` — no colon after the tool name — so the hardening tests do not trip the gate. My first attempt used `\(ERROR\|WARNING\): [A-Za-z]*Sanitizer` and would have failed the nightly on exactly those tests.

Checked against the 609 real `*.asan.log*` files from local sanitizer runs: the only lines matched are `ERROR: LeakSanitizer: detected memory leaks` and `SUMMARY: AddressSanitizer: N byte(s) leaked`, both of which the old pattern also caught, plus nothing from the `failed to allocate` logs.

Keeping `halt_on_error=0` is deliberate now that the scan is real — it lets one run surface every finding instead of only the first. `allocator_may_return_null=1` stays for the reason you give.

Two related things that came out of checking this, both pre-existing and out of scope here:

- `master`'s `sbin/memcheck-summary` is a different, newer implementation with a broken invocation: `grep -q -e "runtime error" -e "Sanitizer" -v "failed to allocate" "$file"` — because `-e` is used, `failed to allocate` is parsed as a *filename*, not a pattern. The master nightly log shows `grep: failed to allocate: No such file or directory` 16 times. Worth its own ticket.
- On arm64, 593 of those 609 local logs contain a genuine `LeakSanitizer` report for redis's own `readQueryFromClient` querybuf (16390 bytes, `redis/src/networking.c`), not module code. x64 CI does not produce it — master's nightly emits only `failed to allocate` logs — but if a sanitizer job is ever run on arm64 it will go red on that, with the old pattern as much as the new one.

tests/flow/tests.sh
AvivDavid23: Settled, and it comes out on the "not real" side — so no `CC=clang` on 2.6, and the comment in #1052 has been reworded (cf19ca5 parent, `.install/install_redis.sh`).

The decisive evidence is that `master` already runs exactly the configuration you describe as fatal, on x64, and is green:

- master's `.install/install_redis.sh` is unchanged in the `make SANITIZER=... / make install` part (the only diff vs 2.6 is `REDIS_REF` derivation from `pack/ramp.yml`), so redis is built by the image's default `cc` — gcc on jammy — with `-fsanitize=address`, no `CC` override.
- master's module is clang-built, same as here (`clang-sanitizer.defs` forces `CLANG=1` when `SAN` is set).
- Nightly run 31906795267 (2026-08-15): `linux-sanitizer / x64-jammy` **success**, 4 × `Total Tests Run: 206, Total Tests Failed: 0`. Same for 08-14, 08-12, 08-11.

So a clang-built module resolving `__asan_*` against gcc's `libasan` is fine at least for redis 7.2/unstable on x64 jammy — the ASan ABI versions line up. 2.6 is left alone.

What is specific to 2.4 is that redis 6.2 has no `ifdef SANITIZER` in `src/Makefile` at all, so `make SANITIZER=address` silently produced an uninstrumented server; that is the part `#1052` actually has to fix, via `REDIS_CFLAGS`/`REDIS_LDFLAGS`/`MALLOC=libc`. `CC=clang LD=clang` stays there because it is what was verified working on that branch, but the comment no longer claims a general gcc/clang incompatibility — it now says the abort was observed with redis 6.2 built by gcc, and points at 8.x/master as the counterexample.

tests/flow/tests.sh
```diff
@@ -159,42 +156,16 @@ setup_rltest() {
...
 #----------------------------------------------------------------------------------------------
 
-setup_clang_sanitizer() {
-	local ignorelist=$ROOT/tests/memcheck/redis.san-ignorelist
-	if ! grep THPIsEnabled $ignorelist &> /dev/null; then
-		echo "fun:THPIsEnabled" >> $ignorelist
-	fi
-
-	# for RediSearch module
-	export RS_GLOBAL_DTORS=1
```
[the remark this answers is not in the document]
AvivDavid23: The premises are all correct, but the conclusion does not hold — I tested it rather than argued it.

**Premises confirmed.** In the jammy sanitizer image (gcc 11.4.0, clang 14.0.0):

```
redis-server links:              libasan.so.6   <- gcc's runtime
redisbloom.so built by:          clang
redisbloom.so links asan rt:     NONE
```

So this is exactly a clang-instrumented `.so` with no ASan runtime of its own, loaded into a gcc-ASan Redis.

**"Can abort on load" — does not.** Every ASan symbol the module needs is exported by gcc's runtime, and both sides agree on the ABI version:

```
module needs 29 __asan_/__sanitizer_ symbols
gcc libasan.so.6.0.0 exports  910
UNRESOLVED: none
module wants:  __asan_version_mismatch_check_v8
gcc provides:  __asan_version_mismatch_check_v8
```

The module loads and serves normally (`bf.add k a` / `bf.exists k a` → 1 / 1) under that server.

**"Or leave the job green without reliable module coverage" — this is the load-bearing claim, and it is false.** Built a stand-in for the module — compiled by clang with `-fsanitize=address`, linked *without* it, so it carries no runtime, exactly like `redisbloom.so` — with a deliberate 1-byte heap overflow, and `dlopen`'d it from a host built by `gcc -fsanitize=address`:

```c
__attribute__((visibility("default"))) int boom(void) {
    volatile char *p = (char *)malloc(16);
    memset((void *)p, 'x', 16);
    volatile int i = 16;
    p[i] = 'y';              /* heap-buffer-overflow, 1 byte past end */
    ...
}
```

```
clang -O1 -g -fPIC -fsanitize=address -fno-omit-frame-pointer -c mod.c -o mod.o
clang -shared -o mod.so mod.o          # no -fsanitize at link
gcc   -O1 -g -fsanitize=address -o host host.c -ldl
```

x86_64 (the CI arch):

```
==1451==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x502000000080
WRITE of size 1 at 0x502000000080 thread T0
    #0 0x7ffffbf1b318 in boom /work/mod.c:9
    #1 0x555555555288 in main /work/host64.c:10
0x502000000080 is located 0 bytes to the right of 16-byte region [0x502000000070,0x502000000080)
allocated by thread T0 here:
    #0 0x7ffffee89887 in __interceptor_malloc asan_malloc_linux.cpp:145
```

aarch64, same result, plus the full shadow map and `SUMMARY: AddressSanitizer: heap-buffer-overflow /work/mod.c:9 in boom`. Correct source line, correct redzone attribution, correct allocation site. Instrumentation in the clang-built object is fully live against gcc's runtime.

**Why the silent-degradation mode cannot occur.** Undefined `__asan_*` symbols are all-or-nothing at `dlopen`: either every one binds, giving full instrumentation, or `dlopen` fails, Redis refuses the module, and every test class errors in `__init__`. There is no code path that yields partial instrumentation. A runtime mismatch here is loud, never green.

**Not introduced by this PR.** 8.4, 8.6, 8.8 and master all pair a gcc-built Redis (`make SANITIZER=address`, no `CC` override) with a clang-built module. master's `linux-sanitizer / x64-jammy` is green — nightly run 31906795267, 4 × `Total Tests Run: 206, Total Tests Failed: 0`, and likewise on 08-14, 08-12 and 08-11. This PR aligns 2.6 with that, it does not invent it.

**Where the concern is not baseless.** The ASan ABI version *is* a real compiler/runtime contract. If jammy's gcc and clang ever land on different `__asan_version_mismatch_check_vN`, `dlopen` breaks and the failure reads as nonsense rather than as a toolchain mismatch. Nothing pins that pairing today. That is worth its own ticket — either build Redis with clang wherever `SANITIZER` is set (which is what #1052 has to do anyway for its redis 6.2 pin) or assert the pairing at build time. It is not a reason to hold 2.6, which is only being brought in line with the branches already running this configuration green.

gabsow (approved):

PR #1052, opened 2026-08-16
2.4 ci: fix nightly sanitizer and valgrind jobs

AvivDavid23:
The `linux-sanitizer` and `linux-valgrind` nightly jobs have both been failing on
this branch for a while ([sanitizer](https://github.com/RedisBloom/RedisBloom/actions/runs/30031037416/job/89287450606),
[valgrind](https://github.com/RedisBloom/RedisBloom/actions/runs/30031037416/job/89287443697)).
None of the failures were module bugs. Companion to #1051 (2.6).

## `tests/flow/tests.sh`

The sanitizer job aborted with `Cannot find redis-server-asan-6.2`:
`setup_redis_server` looked for that exact name, but `getredis --suffix asan` installs the
binary as plain `redis-server-asan`, so it never matched. The Docker image already builds
redis with `SANITIZER=address`, so rebuilding redis here was both redundant and broken —
dropped, and `redis-server` from `PATH` is used instead, as on 8.4. `ASAN_OPTIONS` gains
`allocator_may_return_null=1` so the oversized-allocation hardening tests do not abort the run.

The valgrind job crashed the server in
`test_cuckoo.py:test_number_of_buckets_does_not_overflow`. That test loads a crafted CF
header declaring 2^32 buckets, so `CFHeader_Load` callocs 4 GiB. `getredis --valgrind`
builds redis with `MALLOC=libc`, which routes that allocation through valgrind's own
allocator and makes valgrind run out of memory. Using the image's jemalloc redis keeps the
allocation out of valgrind's allocator; verified the exact `CF.LOADCHUNK` + `CF.INSERT`
sequence now passes under valgrind with an empty valgrind log and the server alive.

Also dropped `[[ -n $SAN ]] && QUICK=1` so sanitizer runs cover every topology, as on
8.4/master. In practice that line only ever suppressed the `oss-cluster` variant: the
Makefile exports `GEN`/`SLAVES`/`AOF`, so the `${VAR:-0}` defaults in the `QUICK` branch
could not turn those off, but it exports `OSS_CLUSTER` while this script reads `CLUSTER`,
leaving `CLUSTER` unset and defaulted to `0`.

## `.install/install_redis.sh`

Dropping the `getredis` path exposed two further problems behind it, both specific to this
branch's redis 6.2 pin:

1. **Redis only grew Makefile `SANITIZER` support in 7.0.** On 6.2 the flag is silently
   ignored (`grep -c SANITIZER redis/src/Makefile` → `0`), so the image shipped an
   uninstrumented server and the sanitized module could not even be `dlopen`'d:
   `undefined symbol: __asan_option_detect_stack_use_after_return`. The equivalent flags are
   now passed by hand when `src/Makefile` has no `ifdef SANITIZER`, mirroring what redis 7.x
   does for `SANITIZER=address` (it also forces libc malloc).

2. **Compiler mismatch.** The module is instrumented by clang (readies
   `mk/clang-sanitizer.defs`) while redis was built with gcc. A clang-built shared object links
   no ASan runtime of its own, so it resolves its `__asan_*` symbols against whatever runtime
   redis already loaded; with redis 6.2 built by gcc on jammy, module load aborted with a
   `global-buffer-overflow` in `RM_GetApi`. Redis is now built with `CC=clang LD=clang` whenever
   `SANITIZER` is set, so both sides use the same runtime. This is *not* a general gcc/clang
   incompatibility — see the review follow-up below.

Both `make` invocations get the same arguments so `make install` does not relink an
uninstrumented binary over the sanitized one.

## `tests/flow/{common,test_tdigest,test_topk}.py`

With the sanitizer job actually running, two tests raced the replica-sync BGSAVE under
`--use-slaves` and failed with `Background save already in progress`.
`repl-diskless-sync` defaults to `no` on redis 6.2, so attaching a replica forks a
disk-backed BGSAVE that blocks `SAVE` and `DEBUG RELOAD`; redis 7+ defaults to diskless,
which is why master never had to deal with this. Under a sanitizer the fork is slow enough
that the race is lost every time. Added `wait_for_no_bgsave()`, called before the direct `SAVE`
in `test_save_load`, and a `dump_and_reload()` wrapper that every `dumpAndReload()` call site
now goes through, so all four topologies stay covered rather than narrowing the job's scope.

## Verification

Ran in the `jammy` container locally. Sanitizer suite green across all four topologies:

| topology | result |
|---|---|
| general | 138 run / 0 failed |
| `--use-slaves` | 138 run / 0 failed |
| `--use-aof` | 138 run / 0 failed |
| `--env oss-cluster` | 138 run / 0 failed |

The local valgrind run is **not** a clean pass — see the note below.

## Open item for CI

Local valgrind runs (on **arm64**, whereas CI is x64) hit `TDIGEST.REVRANK` off-by-one
assertion failures in `test_tdigest_revrank` and `test_tdigest_rank_and_revrank`, e.g.
`[15, 14, 13, 10, 7, 2, -1]` vs `[15, 15, 13, 11, 7, 3, -1]`. These are numeric, not memory
errors, and 2.6 shows the same thing. I could not attribute them confidently: `VG=1` also
implies `DEBUG=1` and adds `-D_VALGRIND`, and valgrind's FP emulation differs by
architecture. The x64 valgrind job on this PR should settle whether it is real or an arm64
artifact.

## Review follow-ups

- `halt_on_error=0` had nothing to fall back on here: 2.4 has no `sbin/memcheck-summary`, and
  `$SAN_ARGS` — the only thing that makes RLTest pass `--sanitizer` and write
  `logs/*.asan.log` — was never assigned. A recoverable module-side finding would have printed
  into the server log and left the nightly green. `halt_on_error` is dropped so ASan aborts the
  server on the first report, and `SAN_ARGS` is now set so the report also lands in a file the
  `if: failure()` artifact upload collects. `allocator_may_return_null=1` still covers the
  oversized-allocation hardening tests. This may legitimately turn the job red on its first run
  if such a finding exists on this branch — that is the point.
- `tests.sh` read `$CLUSTER`/`$QUICK`, but the Makefile exports `OSS_CLUSTER`, so
  `make flow-tests OSS_CLUSTER=0` still ran the oss-cluster variant. That was masked by the
  `[[ -n $SAN ]] && QUICK=1` line this PR removes, so both branches now fall back to
  `OSS_CLUSTER`. Same fix on #1051.
- `wait_for_no_bgsave` covered only 2 of the ~9 exposed sites. `common.py` now has a
  `dump_and_reload(env, **kwargs)` wrapper and every `dumpAndReload()` call site goes through
  it (`test_cms.py`, `test_cuckoo.py` ×2, `test_overall.py` ×4, `test_topk.py`).
  `test_topk.py:62` keeps `restart=True` — the restart path does not issue `SAVE`, so it is not
  exposed — and the `retry_with_rdb_reload` aliases are left alone since every use of them is
  commented out.
- The `_halfRoundUp` attempt at the `TDIGEST.REVRANK` divergence has been **reverted** — over
  the reachable domain it is bit-for-bit `round()`, so it could not have changed the result,
  and its `int_part >= 0.0` guard was wrong for `(-1, 0)` where `modf` yields `-0.0`. `src/` is
  now byte-identical to `2.4`; the REVRANK item below stays open and belongs in its own PR.
- The `CC=clang` rationale is narrowed. It stays, because it is what was verified working on
  this branch, but it is not a general gcc/clang incompatibility: `master` builds redis with the
  image's gcc under `SANITIZER=address`, pairs it with the same clang-built module, and its
  `linux-sanitizer / x64-jammy` job is green (run 31906795267 and the three nightlies before
  it). What is genuinely specific to 2.4 is redis 6.2 having no `ifdef SANITIZER` at all, which
  the `REDIS_CFLAGS`/`REDIS_LDFLAGS`/`MALLOC=libc` branch fixes.

---

> [!NOTE]
> **Low Risk**
> Changes are limited to CI install scripts and test infrastructure; no module runtime logic is modified.
> 
> **Overview**
> Fixes failing **linux-sanitizer** and **linux-valgrind** nightly jobs on the 2.4 branch by aligning Redis builds, test harness setup, and persistence-related test timing with how newer branches run CI.
> 
> **Redis install (sanitizer):** When `SANITIZER` is set, `.install/install_redis.sh` now builds Redis with **clang** (matching the clang-instrumented module) and applies sanitizer flags correctly on the pinned **6.2** tree—either via Makefile `SANITIZER` on 7.0+ or manual `REDIS_CFLAGS`/`REDIS_LDFLAGS` plus `MALLOC=libc` when that target is missing. The same `make` arguments are passed to **`make install`** so install does not replace a sanitized binary.
> 
> **Flow test runner:** `tests/flow/tests.sh` stops looking up or building suffix-specific servers (`redis-server-asan-6.2`, valgrind `getredis` builds) and uses **`redis-server` from PATH** (image-built Redis). Sanitizer runs set **`ASAN_OPTIONS`** (including `allocator_may_return_null=1`), wire RLTest **`--sanitizer`**, and no longer force **`QUICK=1`** so all topologies (including OSS cluster via **`OSS_CLUSTER`**) still run.
> 
> **BGSAVE races:** `common.py` adds **`wait_for_no_bgsave`** and **`dump_and_reload`**; RDB reload call sites in CMS, Cuckoo, Bloom, TopK, and TDigest use them so **`SAVE` / `DEBUG RELOAD`** do not hit *Background save already in progress* under slow sanitizer forks with replicas on Redis &lt; 7.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cf19ca5e50edf98b79dae7002b8f42e507a5f0d1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/rm_tdigest.c
+++ b/src/rm_tdigest.c
@@ -75,6 +75,19 @@ double _halfRoundDown(double f) {
     return int_part >= 0.0 ? int_part + 1.0 : int_part - 1.0;
 }
 
+// Counterpart of _halfRoundDown. Not expressed with round(), whose tie behaviour is
+// libc/platform dependent -- under Valgrind on aarch64 an exact .5 rounds to even
+// instead of away from zero, which silently shifts TDIGEST.REVRANK results.
+double _halfRoundUp(double f) {
+    double int_part;
+    double frac_part = modf(f, &int_part);
+
+    if (fabs(frac_part) < 0.5)
+        return int_part;
+
+    return int_part >= 0.0 ? int_part + 1.0 : int_part - 1.0;
+}
+
 /**
  * Command: TDIGEST.CREATE {key} [{compression}]
  *
@@ -478,7 +491,7 @@ static int _TDigest_Rank(RedisModuleCtx *ctx, RedisModuleString *const *argv, in
             const double cdf_val = td_cdf(tdigest, vals[i]);
             const double cdf_val_prior_round = cdf_val * size;
             const double cdf_to_absolute =
-                reverse ? round(cdf_val_prior_round) : _halfRoundDown(cdf_val_prior_round);
+                reverse ? _halfRoundUp(cdf_val_prior_round) : _halfRoundDown(cdf_val_prior_round);
             ranks[i] = reverse ? round(size - cdf_to_absolute) : cdf_to_absolute;
         }
     }
--- a/src/rm_tdigest.c
+++ b/src/rm_tdigest.c
@@ -75,19 +75,6 @@ double _halfRoundDown(double f) {
     return int_part >= 0.0 ? int_part + 1.0 : int_part - 1.0;
 }
 
-// Counterpart of _halfRoundDown. Not expressed with round(), whose tie behaviour is
-// libc/platform dependent -- under Valgrind on aarch64 an exact .5 rounds to even
-// instead of away from zero, which silently shifts TDIGEST.REVRANK results.
-double _halfRoundUp(double f) {
-    double int_part;
-    double frac_part = modf(f, &int_part);
-
-    if (fabs(frac_part) < 0.5)
-        return int_part;
-
-    return int_part >= 0.0 ? int_part + 1.0 : int_part - 1.0;
-}
-
 /**
  * Command: TDIGEST.CREATE {key} [{compression}]
  *
@@ -491,7 +478,7 @@ static int _TDigest_Rank(RedisModuleCtx *ctx, RedisModuleString *const *argv, in
             const double cdf_val = td_cdf(tdigest, vals[i]);
             const double cdf_val_prior_round = cdf_val * size;
             const double cdf_to_absolute =
-                reverse ? _halfRoundUp(cdf_val_prior_round) : _halfRoundDown(cdf_val_prior_round);
+                reverse ? round(cdf_val_prior_round) : _halfRoundDown(cdf_val_prior_round);
             ranks[i] = reverse ? round(size - cdf_to_absolute) : cdf_to_absolute;
         }
     }
--- a/tests/flow/common.py
+++ b/tests/flow/common.py
@@ -68,3 +68,10 @@ def wait_for_no_bgsave(env, timeout=60):
             return
         time.sleep(0.1)
     raise Exception('timed out waiting for BGSAVE to finish')
+
+def dump_and_reload(env, **kwargs):
+    # RLTest's dumpAndReload() issues SAVE before DEBUG RELOAD NOSAVE, so every call site is
+    # exposed to the BGSAVE race above. Route them all through here rather than guarding the
+    # ones that happen to lose it.
+    wait_for_no_bgsave(env)
+    env.dumpAndReload(**kwargs)
--- a/tests/flow/common.py
+++ b/tests/flow/common.py
@@ -1,5 +1,6 @@
 
 import sys
+import time
 import os
 from RLTest import Env, Defaults
 from redis import ResponseError
@@ -55,3 +56,15 @@ def server_version_at_least(env, ver):
 
 def server_version_less_than(env, ver):
     return not server_version_at_least(env, ver)
+
+def wait_for_no_bgsave(env, timeout=60):
+    # Redis < 7 defaults repl-diskless-sync to no, so attaching a replica triggers a
+    # disk-backed BGSAVE. While that fork is alive, SAVE and DEBUG RELOAD fail with
+    # "Background save already in progress" -- and under a sanitizer the fork is slow
+    # enough that tests reliably race it.
+    deadline = time.time() + timeout
+    while time.time() < deadline:
+        if env.execute_command('INFO', 'persistence')['rdb_bgsave_in_progress'] == 0:
+            return
+        time.sleep(0.1)
+    raise Exception('timed out waiting for BGSAVE to finish')
--- a/tests/flow/tests.sh
+++ b/tests/flow/tests.sh
@@ -8,7 +8,6 @@ READIES=$ROOT/deps/readies
 
 cd $HERE
 
-VALGRIND_REDIS_VER=6
 
 #----------------------------------------------------------------------------------------------
 
@@ -43,39 +42,12 @@ help() {
 
 setup_redis_server() {
 	if [[ -n $SAN ]]; then
-		if [[ $SAN == addr || $SAN == address ]]; then
-			REDIS_SERVER=${REDIS_SERVER:-redis-server-asan-6.2}
-			if ! command -v $REDIS_SERVER > /dev/null; then
-				echo Building Redis for clang-asan ...
-				$READIES/bin/getredis --force -v $VALGRIND_REDIS_VER --own-openssl --no-run \
-					--suffix asan --clang-asan --clang-san-blacklist /build/redis.blacklist
-			fi
-
-			export ASAN_OPTIONS=detect_odr_violation=0
-			# :detect_leaks=0
-			# for RLTest
-			export SANITIZER="$SAN"
-
-		elif [[ $SAN == mem || $SAN == memory ]]; then
-			REDIS_SERVER=${REDIS_SERVER:-redis-server-msan-6.2}
-			if ! command -v $REDIS_SERVER > /dev/null; then
-				echo Building Redis for clang-msan ...
-				$READIES/bin/getredis --force -v $VALGRIND_REDIS_VER --no-run --own-openssl \
-					--suffix msan --clang-msan --llvm-dir /opt/llvm-project/build-msan \
-					--clang-san-blacklist /build/redis.blacklist
-			fi
-		fi
-
-	elif [[ $VALGRIND == 1 ]]; then
-		REDIS_SERVER=${REDIS_SERVER:-redis-server-vg}
-		if ! is_command $REDIS_SERVER; then
-			echo Building Redis for Valgrind ...
-			$READIES/bin/getredis -v $VALGRIND_REDIS_VER --valgrind --suffix vg
-		fi
-
-	else
-		REDIS_SERVER=${REDIS_SERVER:-redis-server}
+		export ASAN_OPTIONS="detect_odr_violation=0:halt_on_error=0:detect_leaks=1:allocator_may_return_null=1"
+		# for RLTest
+		export SANITIZER="$SAN"
 	fi
+
+	REDIS_SERVER=${REDIS_SERVER:-redis-server}
 
 	if ! is_command $REDIS_SERVER; then
 		echo "Cannot find $REDIS_SERVER. Aborting."
@@ -206,7 +178,6 @@ OP=""
 [[ $VG == 1 ]] && VALGRIND=1
 [[ $SAN == addr ]] && SAN=address
 [[ $SAN == mem ]] && SAN=memory
-[[ -n $SAN ]] && QUICK=1
 
 if [[ $QUICK == 1 ]]; then
 	GEN=${GEN:-1}
--- a/tests/flow/tests.sh
+++ b/tests/flow/tests.sh
@@ -42,7 +42,15 @@ help() {
 
 setup_redis_server() {
 	if [[ -n $SAN ]]; then
-		export ASAN_OPTIONS="detect_odr_violation=0:halt_on_error=0:detect_leaks=1:allocator_may_return_null=1"
+		# No halt_on_error=0 here: this branch has no ASan log scanner (sbin/memcheck-summary
+		# does not exist on 2.4), so aborting the server on the first report is the only thing
+		# that can turn a module-side finding into a job failure. allocator_may_return_null=1
+		# still keeps the oversized-allocation hardening tests running -- those get NULL back
+		# instead of an ASan report.
+		export ASAN_OPTIONS="detect_odr_violation=0:detect_leaks=1:allocator_may_return_null=1"
+		# tells RLTest to redirect the ASan output into logs/*.asan.log, which the failure
+		# artifact upload collects
+		SAN_ARGS="--sanitizer $SAN"
 		# for RLTest
 		export SANITIZER="$SAN"
 	fi
@@ -179,16 +187,17 @@ OP=""
 [[ $SAN == addr ]] && SAN=address
 [[ $SAN == mem ]] && SAN=memory
 
+# the Makefile exports OSS_CLUSTER, not CLUSTER, so accept both names
 if [[ $QUICK == 1 ]]; then
 	GEN=${GEN:-1}
 	SLAVES=${SLAVES:-0}
 	AOF=${AOF:-0}
-	CLUSTER=${CLUSTER:-0}
+	CLUSTER=${CLUSTER:-${OSS_CLUSTER:-0}}
 else
 	GEN=${GEN:-1}
 	SLAVES=${SLAVES:-1}
 	AOF=${AOF:-1}
-	CLUSTER=${CLUSTER:-1}
+	CLUSTER=${CLUSTER:-${OSS_CLUSTER:-1}}
 fi
 
 MODULE=${MODULE:-$1}
--- a/.install/install_redis.sh
+++ b/.install/install_redis.sh
@@ -17,8 +17,26 @@ cd redis
 git fetch origin ${REDIS_REF}
 git checkout ${REDIS_REF}
 git submodule update --init --recursive
-make SANITIZER=${SANITIZER:-} -j$(nproc)
-make install
+MAKE_ARGS=()
+if [ -n "${SANITIZER}" ]; then
+    # The module is instrumented by clang (readies clang-sanitizer.defs) and a clang-built
+    # shared object links no ASan runtime of its own, so redis must be instrumented by clang
+    # too -- against gcc's libasan the module reports bogus global-buffer-overflows in RM_GetApi.
+    MAKE_ARGS+=(CC=clang LD=clang)
+    if grep -q '^ifdef SANITIZER' src/Makefile; then
+        MAKE_ARGS+=("SANITIZER=${SANITIZER}")
+    else
+        # Redis grew Makefile SANITIZER support in 7.0; older refs ignore it silently and
+        # produce an uninstrumented server the module cannot even dlopen. Pass the equivalent
+        # flags by hand, mirroring what redis 7.x does (it also forces libc malloc).
+        MAKE_ARGS+=(MALLOC=libc)
+        MAKE_ARGS+=("REDIS_CFLAGS=-fsanitize=${SANITIZER} -fno-sanitize-recover=all -fno-omit-frame-pointer")
+        MAKE_ARGS+=("REDIS_LDFLAGS=-fsanitize=${SANITIZER}")
+    fi
+fi
+
+make "${MAKE_ARGS[@]}" -j$(nproc)
+make install "${MAKE_ARGS[@]}"
 cd ..
 
 echo "Redis installed successfully"
--- a/.install/install_redis.sh
+++ b/.install/install_redis.sh
@@ -20,8 +20,11 @@ git submodule update --init --recursive
 MAKE_ARGS=()
 if [ -n "${SANITIZER}" ]; then
     # The module is instrumented by clang (readies clang-sanitizer.defs) and a clang-built
-    # shared object links no ASan runtime of its own, so redis must be instrumented by clang
-    # too -- against gcc's libasan the module reports bogus global-buffer-overflows in RM_GetApi.
+    # shared object links no ASan runtime of its own, so it resolves its __asan_* symbols
+    # against whatever runtime redis already loaded. Build redis with clang too, so both sides
+    # use the same one -- with redis 6.2 built by gcc on jammy, module load aborted with a
+    # global-buffer-overflow in RM_GetApi. (8.x/master pair a gcc-built redis 7.2 with the same
+    # clang module and are green, so this is not a general gcc/clang incompatibility.)
     MAKE_ARGS+=(CC=clang LD=clang)
     if grep -q '^ifdef SANITIZER' src/Makefile; then
         MAKE_ARGS+=("SANITIZER=${SANITIZER}")
--- a/tests/flow/test_cms.py
+++ b/tests/flow/test_cms.py
@@ -29,7 +29,7 @@ class testCMS():
         self.assertEqual(['width', 2000, 'depth', 7, 'count', 5],
                          self.cmd('cms.info', 'cms2'))
         yield 1
-        self.env.dumpAndReload()
+        dump_and_reload(self.env)
         yield 2
         if not VALGRIND:
             if server_version_at_least(self.env, '7.0.0'):
--- a/tests/flow/test_cuckoo.py
+++ b/tests/flow/test_cuckoo.py
@@ -163,7 +163,7 @@ class testCuckoo():
         self.cmd('FLUSHALL')
         self.cmd('CF.RESERVE', 'cf', '1000')
         yield 1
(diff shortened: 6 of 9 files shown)


src/rm_tdigest.c
```diff
@@ -478,7 +491,7 @@ static int _TDigest_Rank(RedisModuleCtx *ctx, RedisModuleString *const *argv, in
             const double cdf_val = td_cdf(tdigest, vals[i]);
             const double cdf_val_prior_round = cdf_val * size;
             const double cdf_to_absolute =
-                reverse ? round(cdf_val_prior_round) : _halfRoundDown(cdf_val_prior_round);
+                reverse ? _halfRoundUp(cdf_val_prior_round) : _halfRoundDown(cdf_val_prior_round);
```
gabsow: For the only inputs this call site can produce, `_halfRoundUp` is bit-for-bit identical to `round()`, so this change is a semantic no-op and cannot fix the reported `TDIGEST.REVRANK` off-by-ones.

`cdf_val` is in `[0,1]` and `size >= 1` (the `size == 0` case returns early), so `cdf_val_prior_round >= 0` always. For `x >= 0`, `modf` gives `int_part = trunc(x)` and `frac` in `[0,1)`:
- `frac < 0.5` -> `trunc(x)`, which is `round(x)`
- `frac >= 0.5` -> `trunc(x) + 1`, which is `round(x)`

I compiled the helper and swept `size` 1..500 x `cdf` in 4001 steps (2M points over the reachable domain): **0 mismatches vs `round()`**.

So the arm64 Valgrind failure quoted in the PR description — `[15, 14, 13, 10, 7, 2, -1]` vs `[15, 15, 13, 11, 7, 3, -1]` — will reproduce identically after this patch. Four positions differ by 1, which cannot all be exact `.5` ties, and even a genuine tie is resolved the same way by both functions.

The comment above also states an incorrect premise: C99 `round()` is specified to round halfway cases away from zero *regardless* of the current FP rounding mode, and Valgrind emulates FP arithmetic but does not substitute its own libm `round()`. Whatever is shifting REVRANK on aarch64 is upstream of this line (most likely `td_cdf`, or the `-D_VALGRIND` / `DEBUG=1` that `VG=1` implies).

Since the open item in the PR description is still unresolved, I would drop this hunk from the CI fix and keep `src/` untouched on a release branch, rather than land a change that has no effect and whose comment records a wrong diagnosis. Same applies to the identical hunk in #1051.

src/rm_tdigest.c
```diff
@@ -75,6 +75,19 @@ double _halfRoundDown(double f) {
...
+// libc/platform dependent -- under Valgrind on aarch64 an exact .5 rounds to even
+// instead of away from zero, which silently shifts TDIGEST.REVRANK results.
+double _halfRoundUp(double f) {
+    double int_part;
+    double frac_part = modf(f, &int_part);
+
+    if (fabs(frac_part) < 0.5)
+        return int_part;
+
+    return int_part >= 0.0 ? int_part + 1.0 : int_part - 1.0;
```
gabsow: Guarding on `int_part` rather than `f` makes this wrong for every negative input in `(-1, 0)`: `modf` returns `int_part = -0.0` there, and `-0.0 >= 0.0` is **true** in IEEE 754, so the function adds 1 instead of subtracting.

Verified by compiling the helper:

```
_halfRoundUp(-0.5) = 1    (round gives -1)
_halfRoundUp(-0.9) = 1    (round gives -1)
```

Not reachable from the current call site (`cdf_val * size` is never negative), but the function is non-static and its comment presents it as a general round-half-away-from-zero utility, so the next caller that can pass a negative gets the wrong sign — and the magnitude is wrong too, not just the sign. `_halfRoundDown` above has the identical defect.

```suggestion
    return f >= 0.0 ? int_part + 1.0 : int_part - 1.0;
```

tests/flow/tests.sh
```diff
@@ -43,40 +42,13 @@ help() {
...
-	elif [[ $VALGRIND == 1 ]]; then
-		REDIS_SERVER=${REDIS_SERVER:-redis-server-vg}
-		if ! is_command $REDIS_SERVER; then
-			echo Building Redis for Valgrind ...
-			$READIES/bin/getredis -v $VALGRIND_REDIS_VER --valgrind --suffix vg
-		fi
-
-	else
-		REDIS_SERVER=${REDIS_SERVER:-redis-server}
+		export ASAN_OPTIONS="detect_odr_violation=0:halt_on_error=0:detect_leaks=1:allocator_may_return_null=1"
```
gabsow: `halt_on_error=0` is new here, and on **2.4** it removes the only thing that could ever fail the sanitizer nightly on an ASan error — this branch has no ASan detection at all.

Two independent reasons nothing catches it:

1. **RLTest is never told a sanitizer is in use.** `$SAN_ARGS` is interpolated into the RLTest config at `tests/flow/tests.sh:114`, but `SAN_ARGS` is never assigned anywhere in this file (grep confirms line 114 is the only occurrence), so `--sanitizer` is not passed. RLTest only redirects ASan output into `*.asan.log` when `--sanitizer` is `addr`/`address` (`RLTest/redis_std.py:259-264`); otherwise the report just lands in the ordinary server log.
2. **There is no log scanner on this branch.** 2.4 has no `sbin/memcheck-summary` — `sbin/` contains only `get-fbinfer`, `getver`, `pack.sh`, `setup`, `system-setup.py`, `upload-artifacts`, `upload-artifacts-s3`. The only summary here is `valgrind_summary`, which greps `logs/*.valgrind.log`.

readies compiles the module with `-fsanitize-recover=all` (`deps/readies/mk/clang-sanitizer.defs`), so module-side findings are recoverable. Concretely: a `heap-use-after-free` in module code now prints its report into the server log, the server keeps running, the tests pass, and the job goes green. Before this line, `halt_on_error` defaulted to `1`, ASan aborted the server on the first report, the test failed, and the nightly went red — which is the entire point of the job.

`allocator_may_return_null=1` already handles the oversized-allocation hardening tests (those get NULL back rather than an ASan report), so `halt_on_error=0` should not be needed for them.

Minimal fix, so the report is at least captured in a scannable file:

```suggestion
		export ASAN_OPTIONS="detect_odr_violation=0:detect_leaks=1:allocator_may_return_null=1"
		SAN_ARGS="--sanitizer $SAN"
		# for RLTest
		export SANITIZER="$SAN"
```

tests/flow/tests.sh
```diff
@@ -206,7 +178,6 @@ OP=""
 [[ $VG == 1 ]] && VALGRIND=1
 [[ $SAN == addr ]] && SAN=address
 [[ $SAN == mem ]] && SAN=memory
-[[ -n $SAN ]] && QUICK=1
 
 if [[ $QUICK == 1 ]]; then
```
gabsow, line 191: With `[[ -n $SAN ]] && QUICK=1` gone, the oss-cluster topology is now unconditional and there is no way left to turn it off from the Makefile — the `OSS_CLUSTER` knob is silently ignored.

`Makefile:279-283` passes `GEN=$(GEN) AOF=$(AOF) SLAVES=$(SLAVES) OSS_CLUSTER=$(OSS_CLUSTER)` to `tests.sh` and never passes `QUICK` or `CLUSTER`. This script reads `$QUICK` and `$CLUSTER`, both always unset when driven from make, so it takes the `else` branch just below and sets `CLUSTER=${CLUSTER:-1}`.

Result: `make flow-tests SAN=addr OSS_CLUSTER=0` still runs `--env oss-cluster`, and so does `make flow-tests SAN=addr QUICK=1` (the Makefile turns `QUICK=1` into `OSS_CLUSTER=0`, which this script never reads). The removed line was masking that in the other direction, which is why it looked like it only ever suppressed the cluster variant — the `OSS_CLUSTER` vs `CLUSTER` mismatch is the actual cause and it is still here.

The PR description already rests on this analysis, so it seems worth fixing the plumbing in the same change — read the name the Makefile actually exports in both branches below:

```bash
	CLUSTER=${CLUSTER:-${OSS_CLUSTER:-0}}   # QUICK branch
	CLUSTER=${CLUSTER:-${OSS_CLUSTER:-1}}   # else branch
```

(or export `CLUSTER` from the Makefile instead — either way the knob starts working.)

tests/flow/common.py
```diff
@@ -55,3 +56,15 @@ def server_version_at_least(env, ver):
 
 def server_version_less_than(env, ver):
     return not server_version_at_least(env, ver)
+
+def wait_for_no_bgsave(env, timeout=60):
```
gabsow, line 60: Good helper, but it is wired into only 2 of the ~9 sites that hit this race, so the suite stays flaky.

RLTest's `dumpAndReload()` non-restart path issues `con.save()` before `DEBUG RELOAD NOSAVE` (`RLTest/redis_std.py:600-603`), so every unguarded `dumpAndReload()` is exposed for exactly the reason your docstring describes. Still unguarded on this branch:

- `test_cms.py:32`
- `test_cuckoo.py:166`, `test_cuckoo.py:195`
- `test_overall.py:100`, `test_overall.py:221`, `test_overall.py:785`, `test_overall.py:804`

And `test_topk.py:62` already carries a hand-rolled workaround for the same thing:

```python
self.env.dumpAndReload(restart=True) # prevent error `Background save already in progress`
```

which suggests the exposure was never specific to the two tests that happened to lose the race in this run. Since RLTest restarts the env per test class, each class gets a fresh replica sync and therefore a fresh BGSAVE to race; which test loses depends on ordering and on how slow the fork is under ASan, so this will resurface as tests are added or reordered.

Covering all of them centrally would be more durable than two call sites — e.g. a `dump_and_reload(env)` wrapper in `common.py` that calls `wait_for_no_bgsave(env)` first, and switching the call sites to it.

One smaller thing on the helper itself: under `--use-slaves`, `dumpAndReload()` calls `save()` on the replica connection too, while this polls only `env.execute_command`, i.e. shard 1's master (`Env.cmd` -> `self.con`). Not currently a problem — a replica taking a disk-backed full sync does not fork — but worth knowing the guard is master-only if it later gets reused for chained-replica or multi-shard setups.

src/rm_tdigest.c
AvivDavid23: Agreed, and dropped — `src/` is now byte-identical to the base branch on both PRs (revert commits e2634ae on #1051, cf19ca5 on #1052). Confirmed with `git diff origin/<base>...HEAD -- src/` returning empty.

Your analysis holds on both points: over the reachable domain (`cdf_val` in `[0,1]`, `size >= 1`) `_halfRoundUp` is `round()`, and the `int_part >= 0.0` guard is wrong for `(-1, 0)` because `modf` yields `-0.0` there. So the patch could not have moved the arm64 REVRANK result, and the comment's premise about C99 `round()` was wrong.

The arm64 Valgrind REVRANK divergence stays open and moves out of the CI fix — it needs to be chased upstream of this line (`td_cdf`, or the `-D_VALGRIND` / `DEBUG=1` that `VG=1` implies), not in the rounding helper.

src/rm_tdigest.c
AvivDavid23: Agreed, and dropped — `src/` is now byte-identical to the base branch on both PRs (revert commits e2634ae on #1051, cf19ca5 on #1052). Confirmed with `git diff origin/<base>...HEAD -- src/` returning empty.

Your analysis holds on both points: over the reachable domain (`cdf_val` in `[0,1]`, `size >= 1`) `_halfRoundUp` is `round()`, and the `int_part >= 0.0` guard is wrong for `(-1, 0)` because `modf` yields `-0.0` there. So the patch could not have moved the arm64 REVRANK result, and the comment's premise about C99 `round()` was wrong.

The arm64 Valgrind REVRANK divergence stays open and moves out of the CI fix — it needs to be chased upstream of this line (`td_cdf`, or the `-D_VALGRIND` / `DEBUG=1` that `VG=1` implies), not in the rounding helper.

tests/flow/tests.sh
AvivDavid23: Confirmed on both counts and fixed (075e894). Verified `sbin/memcheck-summary` does not exist on this branch, and that line 114 is the only occurrence of `SAN_ARGS`.

Took your minimal fix, minus `halt_on_error`:

```bash
export ASAN_OPTIONS="detect_odr_violation=0:detect_leaks=1:allocator_may_return_null=1"
SAN_ARGS="--sanitizer $SAN"
```

Dropping `halt_on_error=0` is the part that restores the gate: with `-fsanitize-recover=all` on the module, `halt_on_error=1` (the default) still exits the server on the first report, the test errors, and the nightly goes red. `SAN_ARGS` is set as well so the report also lands in `logs/*.asan.log` for the `if: failure()` artifact upload — checked that RLTest 0.7.5 does honor it (`redis_std.py:254-259` gates on `sanitizer in ('addr', 'address')` and appends `log_path` to `ASAN_OPTIONS`), and that `setup_redis_server` runs after the `addr` → `address` normalization, so the value passed is accepted.

`allocator_may_return_null=1` stays, for the reason you give.

Note that this may legitimately turn the 2.4 sanitizer nightly red on its first run: if there is a recoverable module-side finding on this branch, it was being swallowed until now. That is the intended outcome, but it is a real possibility rather than a hypothetical.

Contrast with #1051, where `sbin/memcheck-summary` does exist: there `halt_on_error=0` is kept and the scan was widened instead, so one run surfaces every finding.

tests/flow/tests.sh
AvivDavid23: Confirmed and fixed on both branches (075e894 here, 29cad35 on #1051 — the same mismatch exists there, it was just not being uncovered by removing a line).

```bash
CLUSTER=${CLUSTER:-${OSS_CLUSTER:-0}}   # QUICK branch
CLUSTER=${CLUSTER:-${OSS_CLUSTER:-1}}   # else branch
```

Checked the resulting truth table:

| QUICK | CLUSTER | OSS_CLUSTER | → CLUSTER |
|---|---|---|---|
| — | — | — | 1 |
| — | — | 1 (make default) | 1 |
| — | — | 0 | 0 |
| 1 | — | 0 (`make QUICK=1`) | 0 |
| — | 0 | 1 | 0 |

So `make flow-tests OSS_CLUSTER=0` and `make flow-tests QUICK=1` now both suppress the oss-cluster variant, an explicit `CLUSTER` still wins, and the nightly is unchanged — it leaves `OSS_CLUSTER` at the Makefile default of 1, which was already the effective value via `CLUSTER=${CLUSTER:-1}`.

tests/flow/common.py
AvivDavid23: Fixed centrally (075e894). `common.py` now has:

```python
def dump_and_reload(env, **kwargs):
    wait_for_no_bgsave(env)
    env.dumpAndReload(**kwargs)
```

and all the sites you list are routed through it: `test_cms.py:32`, `test_cuckoo.py:166,195`, `test_overall.py:100,221,785,804`, plus `test_topk.py:224` (the two hand-guarded lines collapse into one call).

Two things I left alone deliberately:

- `test_topk.py:62` keeps `dumpAndReload(restart=True)`. The restart path in RLTest does not issue `SAVE`, so it is not exposed to this race; the comment there stays as the record of why it is a restart.
- `test_cuckoo.py:19` and `:391` bind `self.retry_with_rdb_reload = self.env.dumpAndReload`, but that alias is never invoked — every use is commented out (`test_cms.py:112`, `test_cuckoo.py:69,196`, `test_overall.py:22,32,46`). Wrapping a dead alias seemed worse than leaving it.

On the helper itself: acknowledged, the guard polls `env.execute_command`, i.e. shard 1's master only. Left as-is since a replica taking a disk-backed full sync does not fork, but the docstring limitation is real if this is ever reused for chained replicas or multi-shard.

`test_tdigest.py:947` keeps the bare `wait_for_no_bgsave` since it guards a direct `SAVE`, not a `dumpAndReload`.

tests/flow/tests.sh
```diff
@@ -43,40 +42,21 @@ help() {
...
-		REDIS_SERVER=${REDIS_SERVER:-redis-server}
+		# No halt_on_error=0 here: this branch has no ASan log scanner (sbin/memcheck-summary
+		# does not exist on 2.4), so aborting the server on the first report is the only thing
+		# that can turn a module-side finding into a job failure. allocator_may_return_null=1
+		# still keeps the oversized-allocation hardening tests running -- those get NULL back
+		# instead of an ASan report.
+		export ASAN_OPTIONS="detect_odr_violation=0:detect_leaks=1:allocator_may_return_null=1"
+		# tells RLTest to redirect the ASan output into logs/*.asan.log, which the failure
+		# artifact upload collects
+		SAN_ARGS="--sanitizer $SAN"
```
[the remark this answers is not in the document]
AvivDavid23: Not correct — `--sanitizer` does exist in RLTest 0.7.5, and it is accepted from the `@config` file. Checked against the pinned package rather than inferred, and verified live in this branch's own image.

**The option exists on this pin.** From the venv the 2.4 `Dockerfile.jammy` builds (`uv pip install -r requirements_docker.txt` → `RLTest == 0.7.5`, reported by `pip show` as `Name: rltest / Version: 0.7.5`):

```
__main__.py:261   parser.add_argument('--sanitizer', default=None,
                                      help='type of CLang sanitizer (addr|mem)')
__main__.py:412   if self.args.sanitizer:
__main__.py:413       sanitizer = self.args.sanitizer
__main__.py:450   Defaults.sanitizer = sanitizer
```

**The `@config` concern does not apply either.** This is the one thing that could have made the report true: plain `argparse` with `fromfile_prefix_chars` treats an entire line as a single token, so `--sanitizer address` on one line would indeed arrive as one unrecognized argument. RLTest overrides that behaviour:

```python
__main__.py:30    class CustomArgumentParser(argparse.ArgumentParser):
__main__.py:34        def convert_arg_line_to_args(self, line):
                          for arg in shlex.split(line):
                              if not arg.strip():
                                  continue
                              if arg[0] == '#':
                                  break
                              yield arg
__main__.py:102   parser = CustomArgumentParser(fromfile_prefix_chars=RLTest_CONFIG_FILE_PREFIX, ...)
```

`shlex.split` on each line, so `--sanitizer address` becomes two tokens — which is also why the pre-existing `$RLTEST_ARGS` / `$VALGRIND_ARGS` interpolations in that same config file have always worked.

**Verified live in `bloom24-jammy-san`, with a negative control:**

```
$ cat /tmp/cfg
--sanitizer address
--oss-redis-path=/usr/local/bin/redis-server

$ python3 -m RLTest @/tmp/cfg --collect-only
--- exit=0 ---                                        # accepted, no error

$ printf -- '--not-a-real-flag xyz\n' > /tmp/bad
$ python3 -m RLTest @/tmp/bad
__main__.py: error: unrecognized arguments: --not-a-real-flag xyz    # control fires as expected
```

So the sanitizer nightly does execute; it does not exit before any topology runs.

**The `.asan.log` comment is accurate on this pin too:**

```python
redis_std.py:253   def createCmdOSEnv(self, role):
              254       if self.sanitizer != 'addr' and self.sanitizer != 'address':
              255           return self.environ
              256       osenv = self.environ.copy()
              257       san_log = self._getFileName(role, '.asan.log')
              258       asan_options = osenv.get("ASAN_OPTIONS")
              259       osenv["ASAN_OPTIONS"] = "{OPT}:log_path={DIR}".format(OPT=asan_options, DIR=san_log)
```

Note the check accepts `addr` *or* `address` even though the help text says `(addr|mem)`. `tests.sh` normalizes `addr` → `address` (`[[ $SAN == addr ]] && SAN=address`) before `setup_redis_server` assigns `SAN_ARGS`, so the value that reaches RLTest hits that branch and the per-server `ASAN_OPTIONS` gains `log_path=<name>.asan.log`.

For what it is worth, the same mechanism is what produces the `*.asan.log` files that master's sanitizer nightly currently writes, so it is exercised in CI today — just on a newer RLTest pin.

gabsow (approved):

PR #1063, opened 2026-08-30
MOD-17398 CMS Extention

AvivDavid23:
Features

- `CELL_SIZE 1|2|4|8` — `CMS.INITBYDIM` and `CMS.INITBYPROB` accept an optional trailing `CELL_SIZE` argument selecting bytes per counter cell, cutting memory up to 4× for low-count workloads (1000×5 sketch: 5072 B at `CELL_SIZE` 1 vs 40072 B at 8).
- Negative `CMS.INCRBY` — increments may now be negative, decrementing an item's count when every one of its cells holds at least that magnitude.
- `CMS.INFO` reports cell size — the reply grows from 3 to 4 field pairs so clients can discover a sketch's cell width after creation.

Behavior changes

- Over/underflow is atomic — a rejected increment now leaves every cell and the total count untouched, replacing the old behavior of saturating cells to `UINT32_MAX`, advancing the total count, and returning an error.
- `CMS.MERGE` requires a matching cell size across destination and all sources, erroring with CMS: cell size is not equal rather than reinterpreting a source's bytes at the wrong stride.
- The total count is capped at `INT64_MAX` so `CMS.INFO` count cannot exceed what a RESP integer can carry (a row holds width cells each up to `INT64_MAX`).

Fixes

- Overflow detection moved out of band — `rm_cms.c` previously inferred overflow from a returned sentinel of `UINT32_MAX`, which with `CELL_SIZE` 1 would have reported a legitimate count of 255 as an error; `CMS_IncrBy` now returns an explicit `CMSStatus`.
- `LLONG_MIN` increments are rejected at the parse boundary with CMS: invalid increment, since its magnitude (`2⁶³`) exceeds every cell maximum by one and could only ever underflow.

---

> [!NOTE]
> **Medium Risk**
> Changes CMS persistence format, overflow semantics (breaking for clients relying on saturation), and core counter arithmetic paths; extensive new tests mitigate regressions.
> 
> **Overview**
> Extends Count-Min Sketch with **configurable counter width** and safer arithmetic, plus several **behavior changes** clients should know about.
> 
> **`CMS.INITBYDIM` / `CMS.INITBYPROB`** accept optional **`CELL_SIZE 1|2|4|8`** (default remains 4-byte cells). Sketches store counters in variable-width cells via `cellGet`/`cellSet`, which reduces memory for low-count workloads. **`CMS.INFO`** adds a **`cell size`** field; RDB encoding bumps to **version 1** (still loads old dumps as 4-byte cells).
> 
> **`CMS.INCRBY`** now allows **negative increments** (decrement) with explicit **`CMS: INCRBY underflow`** errors; **`LLONG_MIN`** is rejected at parse time. Overflow/underflow handling is **all-or-nothing**: failed updates **roll back** partial row writes and leave the total **`count`** unchanged—replacing the old saturate-to-`UINT32_MAX` behavior. **`CMS_IncrBy`** returns **`CMSStatus`** instead of inferring overflow from a sentinel.
> 
> **`CMS.MERGE`** requires matching **cell size** across sketches. **`RESTORE`/RDB load** runs **`CMS_ValidateLoaded`** to reject crafted payloads with impossible counters or 8-byte cell values above **`INT64_MAX`**. Command metadata in **`commands.json`** and **`cms_info.c`** is updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 985b375edefd26306730637b5c454d670da78e2c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/cmd_info/cms_info.c
+++ b/src/cmd_info/cms_info.c
@@ -26,7 +26,7 @@ static const RedisModuleCommandArg CMS_INCRBY_ARGS[] = {
 
 static const RedisModuleCommandInfo CMS_INCRBY_INFO = {
     .version = REDISMODULE_COMMAND_INFO_VERSION,
-    .summary = "Changes the count of one or more items by increment, which may be negative",
+    .summary = "Increases the count of one or more items by increment, which may be negative",
     .complexity = "O(n) where n is the number of items",
     .since = "2.0.0",
     .arity = -4,
--- a/src/cmd_info/cms_info.c
+++ b/src/cmd_info/cms_info.c
@@ -26,7 +26,7 @@ static const RedisModuleCommandArg CMS_INCRBY_ARGS[] = {
 
 static const RedisModuleCommandInfo CMS_INCRBY_INFO = {
     .version = REDISMODULE_COMMAND_INFO_VERSION,
-    .summary = "Increases the count of one or more items by increment",
+    .summary = "Changes the count of one or more items by increment, which may be negative",
     .complexity = "O(n) where n is the number of items",
     .since = "2.0.0",
     .arity = -4,
@@ -60,7 +60,7 @@ static const RedisModuleCommandInfo CMS_INFO_INFO = {
 };
 
 // ===============================
-// CMS.INITBYDIM key width depth
+// CMS.INITBYDIM key width depth [CELL_SIZE 1|2|4|8]
 // ===============================
 static const RedisModuleCommandKeySpec CMS_INITBYDIM_KEYSPECS[] = {
     {.flags = REDISMODULE_CMD_KEY_RW,
@@ -74,6 +74,16 @@ static const RedisModuleCommandArg CMS_INITBYDIM_ARGS[] = {
     {.name = "key", .type = REDISMODULE_ARG_TYPE_KEY, .key_spec_index = 0},
     {.name = "width", .type = REDISMODULE_ARG_TYPE_INTEGER},
     {.name = "depth", .type = REDISMODULE_ARG_TYPE_INTEGER},
+    {
+        .name = "cellsize",
+        .type = REDISMODULE_ARG_TYPE_BLOCK,
+        .flags = REDISMODULE_CMD_ARG_OPTIONAL,
+        .subargs = (RedisModuleCommandArg[]){{.name = "cell_size",
+                                              .type = REDISMODULE_ARG_TYPE_PURE_TOKEN,
+                                              .token = "CELL_SIZE"},
+                                             {.name = "size", .type = REDISMODULE_ARG_TYPE_INTEGER},
+                                             {0}},
+    },
     {0}};
 
 static const RedisModuleCommandInfo CMS_INITBYDIM_INFO = {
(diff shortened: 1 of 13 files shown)


src/cmd_info/cms_info.c
```diff
@@ -26,7 +26,7 @@ static const RedisModuleCommandArg CMS_INCRBY_ARGS[] = {
 
 static const RedisModuleCommandInfo CMS_INCRBY_INFO = {
     .version = REDISMODULE_COMMAND_INFO_VERSION,
-    .summary = "Increases the count of one or more items by increment",
+    .summary = "Changes the count of one or more items by increment, which may be negative",
```
TalBarYakar: this is a weird summary 
`Adjusts the count of one or more items by the specified amount, which may be positive or negative.`

src/rm_cms.c
```diff
@@ -286,13 +322,15 @@ int CMSKetch_Info(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
...
-    RedisModule_ReplyWithMapOrArray(ctx, 3 * 2, true);
+    RedisModule_ReplyWithMapOrArray(ctx, 4 * 2, true);
     RedisModule_ReplyWithSimpleString(ctx, "width");
     RedisModule_ReplyWithLongLong(ctx, cms->width);
     RedisModule_ReplyWithSimpleString(ctx, "depth");
     RedisModule_ReplyWithLongLong(ctx, cms->depth);
     RedisModule_ReplyWithSimpleString(ctx, "count");
     RedisModule_ReplyWithLongLong(ctx, cms->counter);
+    RedisModule_ReplyWithSimpleString(ctx, "cell size");
+    RedisModule_ReplyWithLongLong(ctx, cms->cellSize);
```
TalBarYakar, line 333: check if this braking changed well doc

src/rm_cms.c
```diff
@@ -159,11 +184,18 @@ int CMSketch_IncrBy(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
...
+            break;
+        case CMS_STATUS_OVERFLOW:
             RedisModule_ReplyWithError(ctx, "CMS: INCRBY overflow");
+            break;
+        case CMS_STATUS_UNDERFLOW:
+            RedisModule_ReplyWithError(ctx, "CMS: INCRBY underflow");
+            break;
         }
     }
     RedisModule_ReplicateVerbatim(ctx);
```
TalBarYakar, line 201: do u call replicte even though every incrby might fail

src/cms.c
```diff
@@ -63,44 +100,97 @@ void CMS_Destroy(CMSketch *cms) {
...
+    assert(count);
 
-    size_t minCount = (size_t)-1;
+    const uint64_t cellMax = CMS_CELL_MAX(cms->cellSize);
+    const uint64_t magnitude = (value < 0) ? -(uint64_t)value : (uint64_t)value;
 
+    // First pass validates the whole operation, so that a rejected increment
+    // leaves the sketch untouched.
     for (size_t i = 0; i < cms->depth; ++i) {
         uint32_t hash = CMS_HASH(item, itemlen, i);
```
TalBarYakar: computing hash twice , why not caching it in stucked array ?

src/rm_cms.c
AvivDavid23: written in the PR desc + tested

src/rm_cms.c
AvivDavid23: Yes, as same fo the rest of the commands.

src/cms.c
```diff
@@ -135,8 +144,9 @@ CMSStatus CMS_IncrBy(CMSketch *cms, const char *item, size_t itemlen, int64_t va
 
     uint64_t minCount = UINT64_MAX;
     for (size_t i = 0; i < cms->depth; ++i) {
-        uint32_t hash = CMS_HASH(item, itemlen, i);
-        size_t loc = (hash % cms->width) + (i * cms->width);
+        const size_t loc = (i < CMS_LOC_SCRATCH)
```
TalBarYakar: this fix is not good enough 
it is a patch

TalBarYakar (approved):

PR #1068, opened 2026-09-23
MOD-18751: waive cluster bus protected mode for the oss-cluster phase

LiranAbir:
## Problem

[redis/redis#15722](https://github.com/redis/redis/pull/15722) added `cluster-bus-port-protected-mode`, and on the **8.12 line it defaults to enabled**. A node started with `cluster-enabled yes` and without `tls-cluster` then refuses to start, because its cluster bus port would be unauthenticated. That is exactly how the `oss-cluster` phase is built, so every shard dies at startup and each test fails at connect:

```
Running tests on OSS cluster:
  Unhandled exception: Error 111 connecting to localhost:6379. Connection refused.
```

The refusal happens while the configuration is being validated, before the server opens its logfile, so the retained logs are empty and say nothing about the cause. Tracked as [MOD-18751](https://redislabs.atlassian.net/browse/MOD-18751) — every `oss-cluster` leg across this repo, RedisBloom and RedisJSON, public and dev forks, has been red on every run since 2026-09-15.

## Change

- Waive the protection for the `oss-cluster` phase.
- Bump the RLTest pin, in both `requirements_docker.txt` and the tests' `requirements.txt`, to a version that has the option.

Waiving is the documented condition, not a shortcut: the bus ports here are bound to localhost on an ephemeral host, and they were equally unauthenticated before the redis change — which only made running them an explicit choice. redis waived it the same way in its own harness in that PR.

## Why it is not passed unconditionally

A redis that does not have the option **rejects the directive and fails to start the same way**, whichever value is given — the name is what is unknown, not the value. So `CLUSTER_BUS_PROTECTED_MODE=` omits it, for anyone running the cluster phase against an older redis locally:

```bash
CLUSTER_BUS_ARGS="--cluster_bus_port_protected_mode ${CLUSTER_BUS_PROTECTED_MODE-no}"
[[ -z ${CLUSTER_BUS_PROTECTED_MODE-no} ]] && CLUSTER_BUS_ARGS=""
```

RLTest takes this as a caller option rather than working it out itself, because nothing it can see separates a build that has the option from one that does not — the 8.12 line reports `8.9.241` in `version.h`, the same as builds from before the change. Versions carrying it: 8.12 and up, plus the 8.2.10 / 8.4.7 / 8.6.7 / 8.8.3 / 8.10.2 backports, where it defaults to *disabled* and the flag is a harmless no-op.

## Untouched

- `tls_cluster` phase — `tls-cluster` authenticates the bus, so protection is already satisfied.
- `existing-env` / RLEC phase — RLTest starts no servers.

## Verification

Built redis at `7a72677e` (`8.9.241`, the 8.12 line) and ran a 3-shard `oss-cluster` through RLTest for each value:

| binary | value | result |
| --- | --- | --- |
| 8.9.241 (8.12 line) | omitted | fails — `Connection refused` (this bug) |
| 8.9.241 (8.12 line) | `no` | **cluster starts** |
| 8.9.241 (8.12 line) | `yes` | fails, as it should — protection enforced |
| 7.2.6 (no option) | omitted | cluster starts |

3/3 shards receive the option when set, 0/3 when omitted.

Note the `build-linux-*` legs install RLTest into the prebuilt ghcr image, whose tag is the redis ref rather than a content hash, so the new pin only takes effect once **Push Docker Images** rebuilds.

[MOD-18751]: https://redislabs.atlassian.net/browse/MOD-18751?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Low Risk**
> Test harness and dependency pins only; no module runtime or production Redis configuration changes.
> 
> **Overview**
> Fixes **oss-cluster** flow tests failing against Redis builds where **cluster-bus-port-protected-mode** defaults on (e.g. 8.12 line): shards no longer refuse to start for an unauthenticated localhost cluster bus.
> 
> **RLTest** is pinned to **0.7.29** in `requirements_docker.txt` and `tests/flow/requirements.txt` so the harness can pass `--cluster_bus_port_protected_mode`. In `tests/flow/tests.sh`, the oss-cluster leg appends that flag (default **`no`**) via `CLUSTER_BUS_ARGS`; setting **`CLUSTER_BUS_PROTECTED_MODE=`** omits the flag so older Redis binaries that reject unknown directives still start.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 96bb43a5e037d3a69498a7c5728ea11e10282c57. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LiranAbir:
**Blocked on the RLTest release.** This pins `RLTest 0.7.29`, which is not on PyPI yet — the tag is on the way. Until it publishes, CI here will fail at dependency install, not on the change itself.

RedisTimeSeries' equivalent PR is git-pinned to the RLTest merge commit, so it can go ahead of this one.

Order that matters: the pin and the `tests.sh` flag must both be in before judging a nightly. RLTest defaults to omitting the option, so the pin alone changes nothing.

LiranAbir:
Capped redis-py below 6 on this branch.

Nothing here pinned the client, and RLTest requires it as `redis = ">=5.0.0"`, so pip installed redis-py 8.1.0. From redis-py 6 on a connection pool carries kwargs that `Redis.__init__` does not accept, so the tests that rebuild a raw client out of an existing pool fail:

```
File "/workspace/tests/flow/test_cms.py", line 529, in test_cell_size_dump_restore
    raw = redis.Redis(**kwargs)
TypeError: Redis.__init__() got an unexpected keyword argument 'himport_registry'
```

That was every `build-linux` leg on both architectures — four tests across `test_cms` and `test_overall`, 238 run, 4 failed. It is [MOD-18341](https://redislabs.atlassian.net/browse/MOD-18341), not new here, but the bump to `RLTest == 0.7.29` is what let the newest client in, so it surfaces on this PR. `redis ~= 5.0.8` is what RedisJSON already pins.

For the record, the RedisTimeSeries counterpart of this waiver also needed a fix for tests that spawn their own cluster-node replicas ([MOD-18906](https://redislabs.atlassian.net/browse/MOD-18906)). RedisBloom needs no equivalent: nothing in this repo starts a `cluster-enabled` redis of its own — there is no `subprocess.Popen` in the test tree at all, and the replicas in `useSlaves` runs are started by RLTest, so the waiver covers them.

[MOD-18341]: https://redislabs.atlassian.net/browse/MOD-18341?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MOD-18906]: https://redislabs.atlassian.net/browse/MOD-18906?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

gabsow (approved):

